PR #49, opened 2015-01-07
Added list of MindForce plugins

TamiasSibiricus:


dereuromark:
Also, please use `PluginName plugin` syntax for adding CakePHP plugins, look at other existent examples.

dereuromark:
Please better use a single PR for each plugin, as you seem to have put ALL in `Templating`, whereas most don't even belong there (e.g. Passengers: "provides basic users management including authentication and authorization").

jadb:
@dereuromark I think he put them together because plugins seem to require each other in order to work (or at least most of them) or maybe I am missing something?

jadb:
In all cases, for whatever it's worth, I agree that they should be spread across the different sections and just some basic documentation (look who's talking!).

dereuromark:
@jadb Dependencies are managed with composer, not via text-grouping here :)

jadb:
@dereuromark :+1:

TamiasSibiricus:
@ADmad Most of this plugins now in early development stage. So documentation is on the way.  I plan to write basic instructions in readme files in continue of next 2 weeks. Anyway when this plugins would be released they will contain more helpful docs. 
Also my englesh not perfect so i can make some mistakes in doc.

@dereuromark No problem. But how I need to name vendorized plugins? As ususal plugin or you have other recommendations.

@jadb Some plugins vendorized for some reasons. For example Editorial plugins group. Core plugin provide unified functions to include plugin in app. It have autoload functions, but it disabled by default. So you always have a choice include editor helper manually or use autoload.
Same things with Garderobe. Someone use Boilerplate css framefork, someone Bootstrap3, etc... So Garderobe Core plugin provide anly autoload and assets url shortening functions. I hope i can describe my ideas in plugins readmes shortly. No good description - no one will use it.

ADmad:
> Most of this plugins now in early development stage. So documentation is on the way.

So let's wait until then. You can propose to have the plugins included one by one once you have at least some basic documentation is ready for each.

TamiasSibiricus:
@ADmad Ok. I agree with your recommendation. So i close this PR until when docs would be ready.

PR #54, opened 2015-01-14
Updated opauth plugin

wouter0100:
For CakePHP 3


dereuromark:
I am fine with this as the original repo owner hasn't migrated his code to Cake3 yet.
So no need to keep the strike though 2.x version when there is a 3.x version.
Once the original owner upgraded, we can add this one as well then.

wouter0100:
Ah, great! Unfortunantly, the creators of opauth seems to be inactive. That's the main reason why I didn't create a pull-request instead of a "new" repo.

Thanks.

dereuromark:
Refs https://github.com/uzyn/cakephp-opauth/issues/46

jadb:
Why have you deleted the repo without notifying us?

PR #66, opened 2015-02-18
Edited README.md to include xcrypto under security

mistaguy:
I have added a security library for encrypting data to and from a cakephp app. This library was tested manually against a cakephp app and a .Net application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,7 @@ Focus is on the specific task.
 ## Security
 *Plugins and information around security, preventing vulnerabilities and protection against XSS and alike.*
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,6 +132,7 @@ Focus is on the specific task.
 ## Security
 *Plugins and information around security, preventing vulnerabilities and protection against XSS and alike.*
 
+- [Xcrypto](https://github.com/Simbacode/xcrypto) -  Allows excyption of request to and from cakephp server using DES,AES,PHP,HASH,RC4,RSA,Rijndael,TripleDES
```
dereuromark: Hm, looking at it, isn't that just a normal PHP lib? What makes this a CakePHP specific tool?

mistaguy: You are right, initial i was using cakephp network namespace and event namespace that i have removed. 
This makes it a pure PHP encryption library

dereuromark: In that case try awesome-php list.

PR #70, opened 2015-02-25
Added category 'Admin Interface and Utilities'

bobmulder:
Added the CakeManager to the category 'Admin Interface and Utilities'

Made a mistake by adding the CakeManager to the master-branche. So, I removed it and added the plugin once to the cake3-branche.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ Focus is on the specific task.
 	- [Skeleton](#skeleton)
 	- [Environment](#environment)
 	- [Debugging](#debugging)
-	- [Admin Interface](#admin-interface-and-utilities)
 	- [Templating](#templating)
 	- [Email](#email)
 	- [Files](#files)
@@ -89,11 +88,6 @@ Focus is on the specific task.
 - ~~[Airbrake plugin](https://github.com/morrislaptop/AirbrakeCake) - Integrate Airbrake for CakePHP exceptions and errors~~
 - ~~[ClearCache plugin](https://github.com/ceeram/clear_cache) - For easily clearing the app cache(s).~~
 - ~~[GraphVizModels plugin](https://github.com/mamchenkov/CakePHP-GraphViz-Models) - Display your model relations graphically.~~
-
-## Admin Interface and Utilities
-*Plugins for your admin-area and managing*
-
-- [CakeManager plugin](https://github.com/cakemanager/cakephp-cakemanager) - Manage your app: User-Management, built in admin-area and many components and behaviors available for your app.
 
 ## Templating
 *Plugins for templating and lexing.*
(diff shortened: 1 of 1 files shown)


jadb:
Please don't add more categories as the ones already there should already cover all scopes.

Furthermore, since the plugin doesn't only solve 'admin interface', adding 'utilities' to the category as well, makes me feel like that category will only fit this particular plugin, so not really useful per se.

Finally and as a last argument, the plugin is advertised as including many components and behaviors - which is why I will suggest the 'Miscellaneous' category for it since all other categories really limit the plugin's scope.

Thanks for making the necessary changes.

bobmulder:
@jadb refering to [CONTRIBUTING](https://github.com/FriendsOfCake/awesome-cakephp/blob/master/CONTRIBUTING.md) sais: "New categories, or improvements to the existing categorisation are welcome."

So, I thought it would be good to add a new one...

I will add the plguin to the 'Miscellaneous' category

README.md
```diff
@@ -337,6 +337,7 @@ Focus is on the specific task.
 - ~~[Jsonrpc plugin](https://github.com/jameswatts/cake-jsonrpc) - Provides server and client implementations of JSON-RPC.~~
 - ~~[Users plugin](https://github.com/CakeDC/users) - For allowing users to register and login manage their profile.~~
 - ~~[MaintenanceMode plugin](https://github.com/awebdeveloper/cakephp-maintenance-mode) - Show a custom template for all requests when in maintainance with optional IP whitelisting.~~
+- [CakeManager plugin](https://github.com/cakemanager/cakephp-cakemanager) - Manage your app: User-Management, built in admin-area and many components and behaviors available for your app.
```
dereuromark: Note that you should add it above the ~~strike through~~ lines, in alphabetical order then, though.

bobmulder: @dereuromark There we go again ;)

jadb:
@bobmulder - I wouldn't have opposed the new category if it did not include 'utilities' and if the plugin's scope was limited to 'admin interface'. Otherwise, I think it's better for the plugin to be in miscellaneous since some people might need some of its components and/or behaviors but not looking for an admin interface.

My suggestion to you would be to split the plugin into multiple ones that fit different categories and put the admin stuff in a separate plugin, because I do think that a category can be added when there is a need for it (where obviously there seems to be a gap - like with admin stuff, not just interface).

Hoping I made myself a little clearer and thanks for the contribution!

bobmulder:
@jadb I thought about splitting the plugins already but was not sure about it.
You told me about the different 'goals' of the plugin, so I agree with you and I think I gonna split them... Nice to discuss with people this way ;)

PR #78, opened 2015-03-14
Add DJJob plugin

deizel:


dereuromark:
The correct plugin name seems to be `CakeDjjob`

deizel:
Your right, I can fix that.. do you prefer [`CakeDjjob`][1] or [`Cake DJJob`][2] as in the readme?

[1]: https://github.com/josegonzalez/cakephp-cake-djjob/blob/master/composer.json#L29
[2]: https://github.com/josegonzalez/cakephp-cake-djjob/blob/master/README.textile#cake-djjob-plugin

dereuromark:
Exactly the plugin name please.

deizel:
Fixed

PR #81, opened 2015-03-23
Update README.md

davidyell:
The Imagine plugin is for 2.x, so needs strikethrough as it will need upgrading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ Focus is on the specific task.
 ## Imagery
 *Plugins for manipulating images.*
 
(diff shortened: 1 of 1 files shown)


dereuromark:
Seems to me that https://github.com/burzum/cakephp-imagine-plugin/tree/3.0 is working just fine already.

davidyell:
Yes, I'd submitted this PR before checking the branches. So if anything it should be updated to link to the 3x branch by default?

dereuromark:
We never do that. Because they will now - with cake3 stable - all switch to master anyway.

PR #84, opened 2015-03-26
Add ShadowTranslate plugin

ADmad:


README.md
```diff
@@ -287,6 +287,7 @@ Focus is on the specific task.
 ## I18n
 *Plugins for I18n (Internationalization) and L10n (Localization)*
 
+- [ShadowTranslate plugin](https://github.com/AD7six/cakephp-shadow-translate) - A plugin with shadow table based replacement for core's Translate behavior.
```
jadb: Alphabetical order ;)

ADmad: Doh :P

ADmad:
Should be also remove Localized plugin from list? I doubt the datasource in there will be updated to 3.x. Even if they are they should be managed independently.

dereuromark:
I hope @lorenzo shares his plans about localization and the plugin around it with us at the open ticket in the corresponding repo first ( https://github.com/cakephp/localized/issues/84 ).

ADmad:
Doh, I said Localized above but I meant Datasources plugins :stuck_out_tongue:

dereuromark:
;) I don't mind having those in the list for visibility in case someone needs it and maybe uses this opportunity to upgrade it (or part of it at least) then.

PR #87, opened 2015-03-27
Added WhosOnline plugin

bobmulder:


jadb:
While I really like what the plugin does, do you have an ETA for it to be standalone?

bobmulder:
Hey @jadb, 

What does ETA mean? The plugin is standalone. You only have to use the right events ;). But I agree it's not documented well yet... Also, I want to make the event-names configurable. The plugin is basically build for the CakeManager Plugin, but is able to run without it. I hope it's clear to you!

Greetz

Bob

jadb:
Thanks for clarifications!

ETA = estimated time of arrival.

I asked because the readme states: "Note that you need the CakeManager Plugin for this plugin!"

Please confirm this plugin is completely decoupled from CakeManager and I'll make sure it's added as I think a couple users will love the feature.

bobmulder:
I will update this stuff in a few days. Feel free to give feedback or contribute! Can you merge this request?

jadb:
I will as soon as it's decoupled from CakeManager. Leaving it open till then and will wait for a sign from you.

dereuromark:
Closing then for now.

bobmulder:
Hi @dereuromark,

Sorry for the late reply. I was working on some other plugins so forgot to finish this PR.
As you see at https://github.com/cakemanager/cakephp-whosonline#component the plugin is standalone, so able to join the cool plugin-list :)

Greetz,

Bob

dereuromark:
Composer still requires it though.
I bet that is what jadb meant.

bobmulder:
Good point. Didn't see that. Fixed it in https://github.com/cakemanager/cakephp-whosonline/commit/5492d4da5254951a651562594711797841fd1ea0

bobmulder:
Thanks. Does @jadb agree with the changes?

dereuromark:
The dependencies are still there, though:

- 'className' => 'CakeManager.Users'
- $this->loadModel('CakeManager.Roles');

it might be best to clean up, have travis and tests running, a little bit more documention and then we can revisit.

bobmulder:
Correction... 

- `$this->loadModel('CakeManager.Roles');`
This is used in the admin-section wich isn't used when the user doesn't use the CakeManager. So the user has nothing to do with that.
About the first point you are right, I will clean it up!

PR #90, opened 2015-04-21
Remove Js Helper from list.

ADmad:
It's a bad practice which we shouldn't promote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,6 @@ Focus is on the specific task.
 - [Chocolate plugin](https://github.com/commercial-hippie/chocolate) - Front-End framework FormHelper extensions.
 - [CommonMark plugin](https://github.com/gourmet/common-mark) - Adds [CommonMark](http://commonmark.org) (markdown) support to models and views.
 - [Feed plugin](https://github.com/dereuromark/cakephp-feed) - Containing an RssView class to easily generate (complex) RSS feeds.
-- [Js plugin](https://github.com/oldskool/cakephp-js) - A library containing JS helper and alike (extracted from CakePHP2.x core).
 - [Liquid plugin](https://github.com/gourmet/liquid) - A plugin to use the Liquid templating language for views.
 - [SocialMeta plugin](https://github.com/gourmet/social-meta) - Adds support for Facebook's OpenGraph and Twitter's Card meta tags.
 - [TwigView plugin](https://github.com/WyriHaximus/TwigView) - A plugin to use the Twig Templating Language for views.


dereuromark:
I wouldn't call it bad practice to wrap JS into PHP.
It is mainly just overkill, but I also use it (e.g. the GoogleMaps class) to simplify it and just pass arrays (or entities) into it.
So I would not like to remove it just because, if the plugin itself is maintained and working.
I don't know that status, though.

dereuromark:
@oldskool Can you give us your opinion?

ADmad:
Yeah i was just pushing my luck with this :)

oldskool:
I haven't actively maintained my plugin, but that's mainly because there are no reported problems with it :wink: . I'm still ready to fix anything if people report problems or if I discover something myself.

As for the background (if anyone cares :smile:): I primarily "revived" the JsHelper functionality because personally I prefer to use it's `buffer()` method to save up all separate JS calls and render them all at once in a single JS block. I combine this with heredoc syntax highlighting in PhpStorm, so you get "almost" raw JS:

image

I'm not sure why it would be considered bad practice, works like a charm for me without generating any `<script>` block floods/overhead. It also drops the necessity of defining `$(document).ready(function() {});` in every single file, since you all bundle them together in the end. In my layout, I just need one `$this->Js->writeBuffer();` and presto, there it all is, nice and compact in a single script block.

In conclusion:

Personally I see many positive aspects in this approach. This is one of the reasons I love the JsHelper and was sad to see it go in CakePHP 3.0. Therefor I created this plugin and it was linked here (I think by @dereuromark as well?). I'm not sure why it should be (actively) discouraged, but I'd love to hear @AD7six's arguments for that.

dereuromark:
I like the abstraction from JS into PHP in some cases, but that is really super hard to get it working for more than a single use case ;) As we found out with the core helper(s).

You linked the wrong AD* by the way :)
Anyway, let's keep it around for a while, maybe you can add some tests etc. And hopefully someone will pick it up and make it a more useful overall-library then.

jadb:
I am for keeping this. Some people just need basic JS while doing RAD and as long as plugin is maintained, I see no problem advertising it (there are other plugins on the list that aren't really code that should be promoted).

ADmad:
@oldskool Js->writeBuffer() is definitely useful. That's the only Js helper feature is used :)  I agree with @dereuromark and @jadb so closing the PR. 

Please just try to add test cases so that users can use it with confidence. Just need to port the test cases from 2.x :)

oldskool:
@ADmad Good point, I'll put that on my to-do.

PR #94, opened 2015-04-29
Update README.md

gwhitcher:
Added CakeBlog.


dereuromark:
The alphabetic order is wrong.

gwhitcher:
Fixed, my bad.

PR #98, opened 2015-05-11
fotografde/cakephp-sms

felixmaier1989:
We've developed a CakePHP library to send SMS out. We basically copied the code for CakeEmail and adjusted it to telephone short messaging requirements.


dereuromark:
I wonder if it should just go into the "misc" section. But it is the wrong branch, you must use the cake2 branch!

PR #109, opened 2015-06-18
Added CakePHP-BsHelpers plugin

WebAndCow:


dereuromark:
Please use the correct name `BsHelpers plugin` as outlined in the contribution section.
And there is a missing `.` at the end of your sentence.

WebAndCow:
Fixed the name and the end of the sentence.
Is it correct ?

dereuromark:
:+1:

dereuromark:
Do you happen to know the main differences between yours and the existing bs3 plugin in the awesome list?

WebAndCow:
Our plugin helps to write easier and faster the code. We have a more "developer-friendly" approach to gain a lot of time during developements repeatedly.
Contrary to Bootscake, we have new helpers (BsHelper and BsFormHelper)  but we keep the CakePHP process to write forms without changing default configuration every time.
Finally, we added too some other librairies to improve the UX. We already used those librairies during our previous projects and they are handpicked to respond to client needs.

PR #112, opened 2015-06-23
Remove "plugin" from link titles.

ADmad:


dereuromark:
That's exactly what I didnt want to remove for the reasons stated earlier. Some resources are not plugins, knowing which are plugins is vital.

:-1:

ADmad:
Would it make you happy if I ensured all plugin descriptions stated they were a plugin?

dereuromark:
Haha, what do you dislike about the fact that plugins have a clear "Name plugin" syntax in the title? The description itself then can be about the actual description.

dereuromark:
Readme still states (and is valid):

> The main advantage of this list compared to plugins.cakephp.org is, that here it can also be a subpart of a plugin - instead of only the whole plugin/repo itself. The grouping might also be more granular. Focus is on the specific task.

As such the suffix should be kept IMO.

ADmad:
I dislike seeing " plugin" repeated in the titles a fuckton amount of time :)

dereuromark:
Just check out https://github.com/FriendsOfCake/awesome-cakephp#migration where this would be less visually clear without proper suffixing.

ADmad:
Then the descriptions should be improved imo.

dereuromark:
The suffix is part of the quick identification. There is also "app", "skeleton", "", ...
And there are none for those who are neither plugin nor skeleton etc.

dereuromark:
@jadb You decide then :)

jadb:
I think we should have separate sections. One for __plugins__ (to avoid what @ADmad dislikes and I understand to a certain extent - just doesn't bother me like it seems to bother him, lol) and one for __other__ type of resources since the others usually aren't what you are looking for when searching for a plugin (skeleton, etc.).

dereuromark:
Currently it is just easier to keep it the way it is, as this outlines the *specific task* as grouping. And the clear pattern of naming helps to identify what it is.
Closing for now.

PR #116, opened 2015-06-28
Added CipherBehavior plugin

adayth:
Plugin link here: https://github.com/adayth/cakephp-cipher-behavior

This is functionality that I developed for a company project. Is tested and published in packagist.

I hope you consider it useful to add to the list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ Focus is on the specific task.
 ## Security
 *Plugins and information around security, preventing vulnerabilities and protection against XSS and alike.*
 
+- [CipherBehavior plugin](https://github.com/adayth/cakephp-cipher-behavior) - Cipher your entities data with this behavior. Encryption is done at PHP level using CakePHP Security class.
 - ~~[Saving model data and security](http://www.dereuromark.de/2010/09/21/saving-model-data-and-security/) - Information on what to be careful about in 2.x when using CRUD forms and saving data.~~
 
 ## Code Analysis


dereuromark:
It would be great to have travis CI testing enabled on it.
Other than that it looks good!

PS: On a personal note I would have probably used the generic name `Cipher` (without the Behavior suffix) for the plugin name, as the behavior is a specific pattern and this plugin might be some day more than only a behavior, maybe also a lib or shell or whatever :) So you are restricting yourself here maybe a little bit.

adayth:
I'll try to add CI testing soon. I need to search how to do it properly. Any ideas or suggestions are welcome.

I understand your point of view on the name, but I prefer to keep Cipher free for a more generic library that anyone could develop. This plugin is intended to be a behavior and nothing else. I think that is more clear to know that a project is only a behavior just by reading its name.

Thank you for the feedback ;)

dereuromark:
Look how other plugins do it. But you basically just have to put the travis file in and enable the github hook for it.

dereuromark:
Looks like a few builts are failing yet: https://travis-ci.org/adayth/cakephp-cipher-behavior

adayth:
Tests are passing right now. Thank you for your help.

P.S.: maybe this is not the right place, but you can point me to any resources about Readme icons to show status of: coveralls, travis and other info.

dereuromark:
Just check the existing plugins (mine for example). There are tons of examples.

dereuromark:
As for the table schema, you can look into creating a plugin migration using the cakephp/migrations plugin.

adayth:
Thanks @dereuromark ! The schema is just an example to show a common use case and binary column use.

dereuromark:
Ah ok ;)

PR #119, opened 2015-07-17
Add UniqueKey plugin

LubosRemplik:


jippi:
Sorry to say, but its not a  good practice to use - will never scale :(

PR #120, opened 2015-07-22
Add the `Muffin/Trash` plugin

jadb:


dereuromark:
If we add the `Muffin/` prefix now they would be needed on others as well I think.

jadb:
Right, @ADmad you think we should?

On another front @dereuromark, I think we should start removing from here the ones that we think do not follow the "right way" of doing things - might help us not have to re-write plugins ourselves.. lolol. I say this because the currently available SoftDelete behavior we advertise on this list is just badly designed IMO.

I agree that we can't play control police or sound biased in any way, BUT, I also wouldn't like to encourage developers to use plugins that aren't *right*. What do you think?

bravo-kernel:
Just my two cents but why not reverse the idea of removing plugins and instead flag plugins that do follow the "right way". Sort of an "core team approved" image, star, whatever. Two benefits:

- freedom of choice is maintained
- could gently guide people towards the "right way" (maybe some guidelines?)

bravo-kernel:
+1 for the prefixes btw

ADmad:
@dereuromark I think we should update others. It would make it easier to differentiate between similarly named plugins.

@jadb I agree about removing plugins from the list. This should be a curated list, not a comprehensive list of all plugins out there. We already have plugins.cakephp.org for that. I don't see any problem being opinionated and removing/rejecting plugins which we think are not up to standard. What do you say @dereuromark ?

jadb:
@bravo-kernel long time!!

bravo-kernel:
<3 cake long time :wink: 

Was totally unaware of http://plugins.cakephp.org/ so... reversing my reversal suggestion, lol

ADmad:
> Was totally unaware of http://plugins.cakephp.org/ so... reversing my reversal suggestion, lol

It is linked in menu on cakephp.org but in general CakePHP team is really bad at advertising / marketing. :stuck_out_tongue:

ADmad:
> freedom of choice is maintained

If there are good quality options available for the same task surely we will list multiple.

> could gently guide people towards the "right way" (maybe some guidelines?)

There's only so much we can do in our free time :)

ionas:
I do also think a "**(recommended)**"  flag would be good!

dereuromark:
Let's update the others than.
Yeah, but too much "recommendation" and stuff might reduce visibility and is sometimes hard to "judge".
As for the removing of bad practice stuff, this is also kind of difficult to find opinions on, unless it's super-clear.. And it might draw some bad blood. But we can give it a try for some of those most outstanding in sense of "wrong".

jhli:
Maybe a rated plugin site makes more sense? I like this as it is.

dereuromark:
@jhli Yes please. This should stay basic. The rest is up to plugins.cakephp.org IMO.

deizel:
There are some guidelines here, if you wish to follow them:

https://github.com/sindresorhus/awesome/blob/master/awesome.md

> Research if the stuff you're including is actually awesome. Put only stuff on the list you or another contributor can personally recommend and rather leave stuff out than include too much.

Then you can wear the cool badge they are working on:

bravo-kernel:
I totally agree with https://github.com/sindresorhus/awesome/blob/master/awesome.md#only-awesome-is-awesome, this one just killed me https://github.com/sindresorhus/awesome/blob/master/awesome.md#pay-attention-to-gramar

jadb:
:+1: for @deizel's suggestion. 

Who wants to start submitting PRs to remove plugins? I think those are good reasons to start removing based on:

- Unmaintained (yes some are already i believe)
- Bad practice or not "the right way".
- Missing documentation, test coverage or both.
- Poorly written documentation (some of mine may apply here, lol).

ADmad:
Let the culling begin!

ADmad:
Merging this. Prefixs can be added to other plugins later.

deizel:
@bravo-kernel PR submitted for "gramar". :joy:

PR #121, opened 2015-07-29
Remove outdated, unmaintained plugins/resources.

ADmad:


ADmad:
@oldskool JsHelper is being removed from the list as I highly doubt it actually works with latest 3.0. For e.g. `JsHelper::__construct()` will generate strict errors as second param is not typehinted as array like core's `Helper::__construct()`. You can submit the plugin again once you have added tests.

@ceeram I removed `Blame` since `Muffin/Footprint` provides the same feature with additional flexibility.

@dereuromark I removed Tools:Slugged as there are two other plugins specifically for slugs generation.

dereuromark:
@ADmad I doubt that those provide similar advanced features as the one from Ad7six that I upgraded here.

dereuromark:
Personally I think the current changes go way in the wrong direction where I would not want this awesome list to go as a full resource of functional code/plugins available - including the ones that are not yet upgraded but easily could. This provides a useful information beyond what any other resource can provide.
I wonder if you shouldn't just start another awesome list of sorts that does what you want and don't mess with this one.. There you could also highlight the "recommended" ones better maybe.

jadb:
IMO, the plugins.cakephp.org site is for non-prejudice stuff - as it aggregates github as well. This should be a curated 'awesome' list - this wasn't the first time I hit something that isn't awesome (either unmaintained or badly written) and I think that users that come to this list are expecting a curation of some sort, it's the 'awesome' list, not the 'cake3 plugins list'. 

I wouldn't oppose creating a new one but then what would we call it since the 'awesome' name is taken?

ADmad:
> I doubt that those provide similar advanced features as the one from Ad7six that I upgraded here.

Okay then I'll add it back. Though I would suggest moving the behavior to it's own repo. Dumping multiple unrelated stuff in a single repo is a bad idea imho. It's not good for maintenance nor does the end user want the cruft of extra classes just to use one.

> including the ones that are not yet upgraded but easily could.

Sure but who's gonna update them? :)  If their authors have not yet upgrade them they either don't want to or don't have the time. Plus removing them from the list doesn't mean they are gone forever. They just have to be resubmitted after they are updated for 3.0.

> Thinking about it it seems that modifying the "dormant" plugins.cakephp.org site would be the smarter move...

Yup but it again boils down to who has the time and motivation to actually do it. "Community" maintained repos have never worked. It's always the core devs who have to do most of the work.

ADmad:
Let plugins.cakephp.org list everything in a unprejudiced manner and let this list be a collection of actually "awesome" things.

ionas:
As a watcher:

I am more happy to see a list that is maintained ONLY in regards to compatibility and ONLY on a by use case. E.g. someone who uses it and is unhappy reports back the issues and then on case by case stuff is being removed. Maintaining that alone will be hard enough and will have holes already. Maintaining it in regards to a certain "quality level" should be something optional by marking those plugins with something like (awesome) or (recommended) or (best-practise).

If you really want to go forward with purging tons of plugins from this list then the first thing this list should do is pointing to plugins.cakephp.org where you get "more plugins".

My last arguments AGAINST cleaning up this list THOROUGHLY are these two:
a.) It discourages new developers pushing stuff into the cake3 eco system which is still quite new
b.) plugins.cakephp.org seem not to have an option to filter between 3.x and 2.x plugins.

As I said, I'd rather see a (best-practise) and even (not-recommended) tag than a removal. Those tags could be applied case by case which makes it a lot easier to maintain for you guys.

As for the plugins page on cakephp.org - it could very well be community maintained if:
* it featured cakephp.org's users
* it had an easy way to filter 2.x from 3.x plugins
* If community members could tag any plugin with which version it worked for them and which version it did not work for them (think thumbs up and down from stackoverflow)
* if you could add comments requesting support from the plugins developers (again, think stackoverflow)
That way the heavy burden of maintaining an almost-complete plugin list is shifted towards the community.

Just my 2 cents.

dereuromark:
> IMO, the plugins.cakephp.org site is for non-prejudice stuff

But @jadb , doesnt the domain "cakephp.org" imply a certain amount of "official"? 
Using that domain implies that it is not just some crap - or we all don't want it to be.
FOC here should be the more informal one :) My 5 cents HEHE

ionas:
[2:30pm] tflight: Is there a way to filter the plugin list at http://plugins.cakephp.org by those that support 3.*?
[2:30pm] ionas: tflight: github cakephp awesome

jadb:
@dereuromark sorry took me long been really busy. I think CakePHP as a framework should not be selecting - but us as FOC have the freedom to do so. Our actions do not represent the CakePHP group as a whole but a fraction of it. I am saddened to see this list slowly grow into another huge one and not *just awesomness*.

dereuromark:
Ok then.

jadb:
@ADmad, please fix conflicts.

dereuromark:
But for the record: I still think it is a bad idea to remove the not yet upgraded (strike through) plugins that would be worthy to be upgraded, as this info would be irrevocably deleted :(

jadb:
I agree on those - we could remove them if they aren't upgraded by end of year though or if they have a good plugin that already replaces them. What do you think @dereuromark?

ADmad:
If the striked out plugins are to be kept then there's no need to resurrect this PR as those were the bulk of the changes. Few other changes can be done using a new PR.

dereuromark:
Closing then for now.

PR #130, opened 2015-09-14
Adds Muffin/Throttle plugin

bravo-kernel:
https://github.com/UseMuffin/Throttle


dereuromark:
I think the "in CakePHP 3" can be removed. It is pretty clear in what context we are.

bravo-kernel:
Very good point, rebased.

PR #137, opened 2015-09-29
Add Settings category and add Settings plugin

bobmulder:


jadb:
Settings fits better into 'environment' category IMO - am not fond of adding more categories.

bobmulder:
> New categories, or improvements to the existing categorisation are welcome.

@jadb I think the settings plugin is different from plugins listed under `Environment`. Thats why I created a new category. If you really prefer I'll push a new commit so it will be placed under `Environment`.

jadb:
@bobmulder please do - environment is good enough. Not so good standing with just 2 links in there and adding another category with just one link... 

Thanks!

bobmulder:
@jadb Thanks, did it ;)

dereuromark:
The order is not correct yet, it should go above all commented out ones.

bobmulder:
@dereuromark I thought I handled the alphabetic-order-rule right :(

dereuromark:
Thats only for the active ones :P

bobmulder:
Haha I'll remember next time ;) Guess it's okay now?

PR #148, opened 2015-11-27
Add CakePHP-Upload 3.x

davidyell:
Now that @savant has done a release, I think we can include this as an option


dereuromark:
Space instead of `-`.

davidyell:
That is the name of the plugin.

dereuromark:
According to their docs (`Plugin::load('Josegonzale/Upload');`) the name is  `'Josegonzale/Upload'`.

dereuromark:
I rather trust composer, though: `Josegonzalez/Upload`

dereuromark:
?

PR #149, opened 2015-12-04
Created an Html Purifier plugin with tests

chrisShick:


chrisShick:
Are there any issues with this to prevent a merge @dereuromark ??

dereuromark:
Not from my site.

Issue #152, opened 2015-12-11
Suggestion: remove non-upgraded repos


bravo-kernel:
May I suggest removing the strike-through repos now that Cake 3 has been out for some time.

dereuromark:
The idea of those is still the same: Provide easy reference to the old ones and make it easier to help upgrading them instead of reinventing the wheel. 
My opinion still stands: Valuable meta data otherwise lost.

bravo-kernel:
If not black and white let's at least sensibly curate the strike-through items, some entries are like needle-to-eye:

- [Books section] is all about 2.x
- [AmazonSdk plugin](https://github.com/mcallisto) repo does not even exist anymore

Not hurting anything is ofcourse purely subjective, for me it makes our list looks less crisp compared to the others.

dereuromark:
Sure

ADmad:
@bravo-kernel I had suggested a earlier too in #121 but as you can see @dereuromark wants to keep them.

dereuromark:
@bravo-kernel I think our lists are still 1000x better than the awesome-php on or similar ones, because ours are actually well maintained. I think that outweighs any of those issues.
But yeah, please lets remove all outdated ones for sure.

ADmad:
> But yeah, please lets remove all outdated ones for sure.

Yay!  @bravo-kernel Would you like to do the honors?

dakota:
Could we perhaps move them into their own subheading ? That way we don't
loose the knowledge, but they no longer clutter the list?

dereuromark:
Then we can remove them alltogether and maybe make a WIKI page containing the TODOs?

dakota:
:+1: for the wiki page.

dereuromark:
Added a wiki start entry. Someone just has to move them over now:
https://github.com/FriendsOfCake/awesome-cakephp/wiki#plugins-not-yet-upgraded-from-2x-to-3x

bravo-kernel:
Awesome (no pun intended), I hereby volunteer to move things over. Will probably not be done before the weekend though.

bravo-kernel:
Steps:
- [x] move all non-upgraded material as-is to [wiki](https://github.com/FriendsOfCake/awesome-cakephp/wiki)
- [ ] remove dead links from wiki
- [ ] remove moved items from the main list
- [ ] curate main list (order, etc)
- [ ] follow-up status questions on the non-upgraded repos (**?!?**)

[...]

PR #155, opened 2015-12-18
Sort categories alphabetically (#152)

bravo-kernel:
Solves single-last curation item described in #152


bravo-kernel:
Triple-checked, should be good to merge. @dereuromark agrees on the sort.

bravo-kernel:
Much obliged, feels like the :syringe: has been removed from my eye

PR #160, opened 2015-12-22
Remove Blame as it has been marked abondoned

ADmad:
https://packagist.org/packages/ceeram/cakephp-blame


Oxicode:
Why. Sí funciona

ADmad:
@Oxicode You should ask that to the plugin author :)

dereuromark:
Well, thats bad because other plugins have been rejected and not been added to the list in favor of Blame.
Any alternatives?

ADmad:
The suggested alternative muffin/footprint is already one the list :)

PR #162, opened 2015-12-30
Add OrgasmicNightmare/cakephp-seo plugin

orgasmicnightmare:
A SEO plugin for cakePHP 3 to manage seo tags with ease. It comes with a Behavior, a Component, a View Cell and the default admin crud actions for each tables.

The behavior helps you to generate some default tags for each model. You can use simple text, pattern, callbacks.
The component inject tags in the view.
The view cell helps you to add or edit tags directly in the admin view of a record.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ Additional lists you might find useful:
 *Search Engine Optimization.*
 
 - [Muffin/Slug plugin](https://github.com/UseMuffin/Slug) - A plugin for generating slugs and finding records by slug. Uses a pluggable architecture which allows using your own slug generator class.
-- [OrgasmicNightmare/cakephp-seo](https://github.com/orgasmicnightmare/cakephp-seo) - Manage seo tags with ease. It comes with a Behavior, a Component, a View Cell and the default admin crud actions for each tables.
+- [Seo plugin](https://github.com/orgasmicnightmare/cakephp-seo) - Manage seo tags with ease. It comes with a Behavior, a Component, a View Cell and the default admin crud actions for each tables.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -275,6 +275,7 @@ Additional lists you might find useful:
 *Search Engine Optimization.*
 
 - [Muffin/Slug plugin](https://github.com/UseMuffin/Slug) - A plugin for generating slugs and finding records by slug. Uses a pluggable architecture which allows using your own slug generator class.
+- [OrgasmicNightmare/cakephp-seo](https://github.com/orgasmicnightmare/cakephp-seo) - Manage seo tags with ease. It comes with a Behavior, a Component, a View Cell and the default admin crud actions for each tables.
```
dereuromark: Please respect the guidelines for the name.

orgasmicnightmare:
Oh my bad, It should be ok.

README.md
orgasmicnightmare: Oh my bad, It should be ok.

README.md
```diff
@@ -275,6 +275,7 @@ Additional lists you might find useful:
 *Search Engine Optimization.*
 
 - [Muffin/Slug plugin](https://github.com/UseMuffin/Slug) - A plugin for generating slugs and finding records by slug. Uses a pluggable architecture which allows using your own slug generator class.
+- [Seo plugin](https://github.com/orgasmicnightmare/cakephp-seo) - Manage seo tags with ease. It comes with a Behavior, a Component, a View Cell and the default admin crud actions for each tables.
```
dereuromark: crud => CRUD imo.
But I wonder if you need to be that specific regarding Behavior, Component etc.
I think the description should not list the exact types of classes used, but the functionality this plugin does.

PR #163, opened 2016-01-03
Add CakeYaml plugin

makallio85:
https://github.com/makallio85/cake-yaml

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,6 @@ Additional lists you might find useful:
 *Useful libraries or tools that don't fit in any of the other categories.*
 
 - [Aura.Intl](https://github.com/auraphp/Aura.Intl) - A powerful I18n library - used in CakePHP3.x core.
-- [Yaml-route](https://github.com/makallio85/yaml-route) - Configure routes with simple YAML files.
 - [Capcake](https://github.com/jadb/capcake) - Deploy CakePHP applications using Capistrano.
 - [Carbon](https://github.com/briannesbitt/Carbon) - A simple DateTime API extension - used in CakePHP3.x core (deprecated soon).
 - [Chronos](https://github.com/cakephp/chronos) - A simple standalone DateTime API extension (successor of Carbon).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -176,6 +176,7 @@ Additional lists you might find useful:
 *Useful libraries or tools that don't fit in any of the other categories.*
 
 - [Aura.Intl](https://github.com/auraphp/Aura.Intl) - A powerful I18n library - used in CakePHP3.x core.
+- [Cake-yaml](https://github.com/makallio85/cake-yaml) - Configure CakePHP 3 routes with simple yaml files.
```
ADmad: "YAML" instead of "yaml" would be better since it's an acronym.

ADmad:
I would prefer if the plugin had tests before it got added to the list.

makallio85:
I added integration test, seems to be running just fine.

dereuromark:
This is a plugin? It should follow the contribution guidelines!
The name is "CakeYaml plugin" then. 
Also the description could be improved. No need to mention CakePHP or the version number as both are clear from context, so maybe just

     Configure routes with simple YAML files.

dereuromark:
It would also better fit in the plugins section then: https://github.com/FriendsOfCake/awesome-cakephp#miscellaneous

makallio85:
Okay, good points. Just one thing. This is actually library and plugin. But not just one. It works only with CakePHP 3 but does not work like a plugin.

So is it plugin or not?

dereuromark:
What do you mean it does not work like a plugin?
You are activating it by using Plugin::load(), are you not? That classifies as plugin usage for me :)

PS: I also made the PR message a little bit less vague.

makallio85:
Okay, then it is not plugin. It is not loaded like that.

makallio85:
Then I'll change:
- description to "Configure routes with simple YAML files."
- package name to yaml-route

Thats is?

dereuromark:
Ah I see, ok sonds good :)

makallio85:
I made those changes. It should fine now.

README.md
```diff
@@ -176,6 +176,7 @@ Additional lists you might find useful:
 *Useful libraries or tools that don't fit in any of the other categories.*
 
 - [Aura.Intl](https://github.com/auraphp/Aura.Intl) - A powerful I18n library - used in CakePHP3.x core.
+- [Yaml-route](https://github.com/makallio85/yaml-route) - Configure routes with simple YAML files.
```
dereuromark: YamlRoute?

makallio85:
YamlRoute it is. Ordered list with update also.

dereuromark:
Cool, thx!

PR #169, opened 2016-01-19
Add webservices plugin

Marlinc:
I've added the https://github.com/usemuffin/webservice plugin which now contains a ORM like interface to work with webservices. I'm not sure if this it the right section, if not this can me moved to another section.


dereuromark:
Fullstop missing at the end.
But looks good.

Marlinc:
There you go!

bravo-kernel:
`An ORM like interface` instead of `A ORM like interface` IMHO

Marlinc:
Good one, it has been changed. I've squashed the commits as well.

PR #178, opened 2016-02-03
AssetCompressImproved plugin and JadeView plugin

clthck:
Add 2 new plugins to the list

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Additional lists you might find useful:
 *Tools for managing, compressing and minifying website assets.*
 
 - [AssetCompress plugin](https://github.com/markstory/asset_compress) - A complete asset manager for CakePHP.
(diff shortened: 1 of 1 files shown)


dereuromark:
Ideally, each addition is an own PR.
Now the one plugin cannot be added because of discussion about the other (new) one.

Why do you need to hard-fork the AssetCompress one?
Wasn't it possible to PR the functionality? As this is an activiely maintained (de facto standard) plugin.

clthck:
I tried to suggest my fix with a pull request, but @markstory didn't want to accept my update. So I decided to make a new one.

clthck:
You can check our conversations [here](https://github.com/markstory/mini-asset/pull/18)

jadb:
I would split this PR in 2.

Also, I would not add a new AssetCompress plugin and mix-up the community about which to use when it only adds a feature for a limited scope of use cases. 

Instead, I suggest we link to it next to AssetCompress and update its README to point to the PR and/or describe what it adds to the original one so people can decide at a glance what to use. 

After all, and unless you keep rebasing from @markstory's plugin (which is the official one we promote to people), users will run into problems - so better they know what advantages they are getting by using it and judge if they want to or not.

PR #181, opened 2016-02-05
Added RelatedContent plugin

aiphee:
I created this plugin for managing related content.
It probably needs someone to test it first, since i'm the only one using it. 

I'm also not sure how do i write tests for behavior.

Examples of usage:
- Article has "read more" for other article
-  News has "read more" for other news or pages
- Article has related galleries

Plugin has two parts, one part is caching, and the second uses the cache.

It is my first plugin for PHP framework ever, so i embrace constructive criticism.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ Additional lists you might find useful:
 - [PersistRelatedData plugin](https://github.com/riesenia/persist-related-data) - Behavior for persisting selected fields of related models.
 - [RowLocker plugin](https://github.com/lorenzo/row-locker) - Exclusive locks for rows in your tables.
 - [Webservices ORM plugin](https://github.com/usemuffin/webservice) - An ORM like interface for webservices.
+- [RelatedContent plugin](https://github.com/aiphee/CakePHP3-related-content-plugin) - Behaviors and element for managing many to many related entities from different models. 
 
 ## PDF
 *Plugins and software for working with PDF files.*


dereuromark:
A few tips:

- Please add tests, so you can have coverage buids for free for all PHP versions and avoid regressions.
- `CakePHP3-related-content-plugin` as github name is not a good choice, because you lock down to this major version and changing it later will break URLs. better to stick to the recommendations of this repo (see the Contritbution notes): `cakephp-related-content` etc. Your readme should point out which cakephp framework version your code is working with.
- If you plan on using the former Cake(2) CS, you should be using [PSR-2-R sniffer tool](https://github.com/php-fig-rectified/psr2r-sniffer) to fix and assert the style. Currently you have a strange indentation that is invalid for old and new style guide (and also inconsistent to the rest of the coding world).

aiphee:
Ok thanks :)
I just have two questions.

 - How do i write test for behavior? I need to mock method, which searches tables with attached behavior, so it returns what i give it. But static methods cant be mocked, and as i understand, not even proper methods in behavior. 
 - How important is same code style? I have no problem switching from tabs to spaces, but for example, i like to have my assignments with aligned "=", which is not correct by CakePHP standarts

I chose GPL because i`m writing this plugin partly for company i work in, and as i understand, GPL is little more restrictive about resharing contributions.

dereuromark:
Please take a look at the existing plugins and their behaviors.
There are tons of tests that show how you can use fixtures and no mocking required then :)

You should at least be consistent. I would recommend either PSR2 or PSR2R.
Both should come with a sniffer for what you mean, if not please help to provide one so we can improve the style in general.

OK, was really just wondering about the GPL part.
Usually we use MIT which works just fine for open source code.

aiphee:
I fixed some PSR2R errors but i cant get through these errors: 
```
 42 | ERROR | [x] Double space found
 60 | ERROR | [x] Table => \Cake\ORM\Table
- // -
```

I dont have to align `=` in assigns if necessery (i cant disable it since same rule also checks for mixed tabs and spaces. But i no idea how should i write this `/** @var Table $table */`, it is autocomplete helper for IDE.

I tried to write tests with Fixtures, but i really need different input from method `getAttachedTables`, i even tried to make it non-static, but cant mock it anyway.
That method fetches Table files and searches in text if they have `InRelatedIndex` behavior, so when i test it, it naturally return nothing, while i need it to return my Fixture tables.

dereuromark:
I think it would be best to join IRC channel for further discussion about your testing issues.
Shouldn't be part of this PR and discussion.

PR #182, opened 2016-02-06
Update README.md

clthck:
Add a new entry in Skeleton section


bobmulder:
I think the implemented design should be placed in a theme? That would make the use much easier because people will use cake's default skeleton mostly...

PR #184, opened 2016-02-09
Added FractalTransformerView plugin

andrej-griniuk:
Hi there,

this is my first public CakePHP plugin, so apologies in advance if I'm missing something.

Cheers,
Andrej

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ Additional lists you might find useful:
 *Plugins and web tools for developing REST-ful APIs.*
 
 - [Cors plugin](https://github.com/snelg/cakephp-cors) - A lightweight plugin for adding CORS headers to specified endpoints.
+- [FractalTransformerView plugin](https://github.com/andrej-griniuk/cakephp-fractal-transformer-view) - A plugin which allows using [Fractal transformers](http://fractal.thephpleague.com/transformers/) for your API output.
 - [Swagger plugin](https://github.com/alt3/cakephp-swagger) - Swagger plugin for documenting your CakePHP 3.x APIs.
 
 ## Search


jadb:
Nice one :+1:

jadb:
@andrej-griniuk would you mind opening a PR on Fractal to promote it also there so people looking to use it with Cake find out about it?

ADmad:
Nice one indeed.

bravo-kernel:
Does this support reverse transformations as well? Presenting transformed/standardized API data to clients is awesome but allowing them to send it in the same format would be :100:.

bobmulder:
Totally agree with that @bravo-kernel, unfortunately Fractal doesn't support the reversed part. I am working on an API plugin which supports the Transformer stuff including the reversed part...

bravo-kernel:
I'm no expert but if format A can be transformed to format B surely the other way around would be possible too? I :+1: joined efforts in a single plugin/repo.

bravo-kernel:
Shamelessly including @josegonzalez since it's inspired on his blog and we discussed this earlier on irc.

josegonzalez:
Heh, maybe could have saved some work :P https://github.com/josegonzalez/cakephp-fractal-entities

bobmulder:
:+1: I like it!

bravo-kernel:
FWIW I think this could be the most popular and most commonly used plugin next to CRUD if it would:

- support transforming incoming data to Cake format
- implement some sort of "strategies" like the [Obfuscate plugin](https://github.com/UseMuffin/Obfuscate). Imagine instant drop-in support for json-api, json-hal, etc. :dancer: 

Would like to discuss if this is possible since I have need of it.

andrej-griniuk:
First of all, thanks everyone. Glad my plugin caused this discussion.

This plugin, as it's name refers to, meant to provide Fractal library support and only handle the output part. As it was mentioned above, unfortunately Fractal doesn't support reverse transformation.

@bobmulder and @bravo-kernel what you are talking about is interesting and I might be able to contribute. However, I don't own this problem so it would be difficult for me to define the requirements and propose a solution.

@jadb happy to, but doesn't look like they have any integrations page, so not sure where I should open a PR?

@josegonzalez it did save indeed! (as it's mentioned in Credits). By the way, your CakePHP Advent articles are always great!

jadb:
@andrej-griniuk - on their README, create a new section if there isn't.

I've done that in the past for league's libraries [here](https://github.com/thephpleague/factory-muffin/#integrations) and [here](https://github.com/thephpleague/commonmark#integrations) for example.

bravo-kernel:
@bobmulder please :flashlight: when you get your plugin up and running and/or need help.

bobmulder:
I've done some research about the imput part of the fractal-library. Under one of the many issues and comments I want to show the following one: https://github.com/thephpleague/fractal/issues/65#issuecomment-46082208

On the other hand I disagree with that, but at the end I have to admit we want to use the library for something it's not meant for. As the main-page says: 

> Create a “barrier” between source data and output, so schema changes do not affect users

So, we need to think about something different to handle input data. 

@bravo-kernel I managed to add the transformers as view layer of the API to my plugin. The difficult thing now is to handle input. I will try to put my code online this week! It's a bit messy... :$

PR #186, opened 2016-02-15
Added TwoFactorAuth plugin

andrej-griniuk:
https://github.com/andrej-griniuk/cakephp-two-factor-auth


dereuromark:
You can start with "Allows two ..." As the context makes it clear it is a plugin.

:+1:

andrej-griniuk:
@dereuromark sure, updated.

PR #187, opened 2016-02-15
Adding JsonApi plugin

josbeir:
https://github.com/josbeir/cakephp-json-api

### DIFF
--- a/README.md
+++ b/README.md
@@ -263,6 +263,7 @@ Additional lists you might find useful:
 
 - [Cors plugin](https://github.com/snelg/cakephp-cors) - A lightweight plugin for adding CORS headers to specified endpoints.
 - [FractalTransformerView plugin](https://github.com/andrej-griniuk/cakephp-fractal-transformer-view) - A plugin which allows using [Fractal transformers](http://fractal.thephpleague.com/transformers/) for your API output.
+- [JsonApi plugin](https://github.com/josbeir/cakephp-json-api) - Implements the [JSON API](http://jsonapi.org/) format.
 - [Swagger plugin](https://github.com/alt3/cakephp-swagger) - Swagger plugin for documenting your CakePHP 3.x APIs.
 
 ## Search


README.md
```diff
@@ -264,6 +264,7 @@ Additional lists you might find useful:
 - [Cors plugin](https://github.com/snelg/cakephp-cors) - A lightweight plugin for adding CORS headers to specified endpoints.
 - [FractalTransformerView plugin](https://github.com/andrej-griniuk/cakephp-fractal-transformer-view) - A plugin which allows using [Fractal transformers](http://fractal.thephpleague.com/transformers/) for your API output.
 - [Swagger plugin](https://github.com/alt3/cakephp-swagger) - Swagger plugin for documenting your CakePHP 3.x APIs.
+- [JsonApi plugin](https://github.com/josbeir/cakephp-json-api) - A plugin that implements the [JSON API](http://jsonapi.org/) format.
```
dereuromark: You can start with `Implements ...`.
That it is a plugin can be seen from the beginning of that row.
Please also respect the alphabetical order.

josbeir:
updated :-)

jadb:
I am not sure we should promote the way the plugin does things. Accessing the request/response from the model layer? Something isn't right.

Also, maybe better re-submit when it is more complete? It says there is still 'lots to do'.

josbeir:
The schema file kind of acts as the view of an entity. It has nothing to do with the model.

ADmad:
`EntitySchema` schema looks like a view transformer. So perhaps it should live under `View` namespace?  

I don't quite understand the purpose of `EntitySchema::__get()`. Why does entityschema need to access a View class property?

Also I am pretty sure your codecoverage config is wrong. With [this]https://github.com/josbeir/cakephp-json-api/blob/master/phpunit.xml#L42) you are testing coverage of the test files (which will obviously give you 100% coverage) instead of checking code coverage of the source files under `/src/`.

ADmad:
> I don't quite understand the purpose of EntitySchema::__get(). Why does entityschema need to access a View class property?

The readme examples answers my question :)  It's to allow this in entityschema:

```php
            'helper_link' => $this->Url->build(['action' => 'view']) // view helper
```

josbeir:
@ADmad thanks for pointing that out, didn't even notice :-) It's fixed now.

About the namespace, yeah i was struggling with that because it also didn't look right to put then in the View namespace. But i guess you are right, i'll update this in the future..

josbeir:
Made a bunch of improvements and updated the documentation
Removed the component because it was just plain useless :-)

dereuromark:
I guess we can merge this.

PR #199, opened 2016-04-10
added skie/PhpStorm link

skie:
https://github.com/skie/PhpStorm

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,6 @@ Additional lists you might find useful:
 	- [Third Party APIs](#third-party-apis)
 - [Software](#software)
 	- [Development Environment](#development-environment)
-	- [IDE Extensions](#ide-extensions)
 - [Web Applications](#web-applications)
 	- [CMS and applications built on CakePHP](#cms-and-applications-built-on-cakephp)
 	- [Demo](#demo)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -365,6 +366,11 @@ Additional lists you might find useful:
 - [Puppet](https://puppetlabs.com/) - A server automation framework and application.
 - [Vagrant](https://www.vagrantup.com/) - A portable development environment utility.
 
+## IDE  extensions
+*Software and tools to integrate with cakephp specific features.*
+
+- [Command line tool console gneration plugin for PhpStorm](https://github.com/skie/PhpStorm) - Builds xml file to use autocomplete for cakephp console from PhpStorm IDE.
```
dereuromark: `XML` and `CakePHP`

dereuromark: Please follow the naming scheme as per Contribution.md

bravo-kernel:
Missing e in `gneration`

dereuromark:
IMO this does not need its own category, putting it into the existing "Development Env" one would totally suffice.

README.md
jadb: For description, how about:

```
[PhpStorm plugin](...) - CakePHP console auto-completion for the PhpStorm IDE
```

skie:
There is plugin usage exampleexample

jadb:
@skie nice! 

Would you mind just bumping it a little up in the list under that category so it's alphabetically sorted?

dereuromark:
The word Console looks wrong to me in that sentence

jadb:
@dereuromark why? What do you suggest?

skie:
The tool name is "Command line Tools Console"
So console is correct here.

dereuromark:
The specifics and that it is a shell is irrelevant here IMO, so it should read what it does:

    CakePHP auto-completion support for the PhpStorm IDE.

or sth like that.

jadb:
@dereuromark I will have to disagree.

I see it provides shell tasks not auto-completion of code snippets (which is what your description makes me think of).

dereuromark:
Then I misread the plugin itself, so it does not help with autocompletion of the code, but the console commands there. Well, then I take my last comment back,of course. Sry for that!

    CakePHP auto-completion support for console command in PhpStorm IDE.

Is that clearer maybe?

skie:
Final update added

dereuromark:
Truly awesome, thank you!

PR #201, opened 2016-04-20
Add DataTables plugin

ypnos-web:
https://github.com/ypnos-web/cakephp-datatables

The plugin consists of a Component and a Helper to easily use jQuery DataTables in a CakePHP app.
The component provides a finder wrapper that processes AJAX requests by DataTables.
The Helper provides an easy method of setting up the table in a template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,7 @@ Additional lists you might find useful:
 - [CakeAdmin plugin](https://github.com/cakemanager/cakephp-cakeadmin) - A non-stable user management plugin with a built-in admin area.
 - [CurrencyConverter plugin](https://github.com/AlessandroMinoccheri/cakephp-currency-converter) - A plugin to convert currency into another one.
 - [Dashboard plugin](https://github.com/gourmet/dashboard) - Build beautiful dashboards for your cakes.
+- [DataTables plugin](https://github.com/ypnos-web/cakephp-datatables) - Create dynamic tables with jQuery DataTables and server-side processing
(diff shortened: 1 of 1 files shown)


dereuromark:
I'd say some tests would be nice to ensure no regressions can happen.

ypnos-web:
Yes, I agree.
Right now I finished an initial stable version that is pretty feature complete.
Tests come next before I add new code to master for the obvious reason.
You can wait for the tests. From my perspective I think it also makes sense not to wait hence the PR.

ADmad:
We'll wait for tests :)

dereuromark:
@ypnos-web Any update?

ayman-alkom:
what about accepting the plugins witch not have tests, by adding a note to the end of it's name.
just an idea

dereuromark:
That's not acceptable IMO. The tests help to see early on if the plugin is still compatible with the future minors and majors, making it easier to avoid regressions.
We owe that to the people deciding for a plugin and starting to use it.
This list is supposed to outline awesome plugins - and that is an important piece of this :)
Please help us keeping this high standard of quality, and making it easy for people to trust into this awesome list.

ayman-alkom:
you are right

ypnos-web:
The status is that I have plans for the plugin, including the tests, but it will take until summer before I get to it. Until then status quo (only bugfixes, tests still missing).

I fully agree with having a strict policy for all entries.

ADmad:
Closing the PR for now in that case. It's already been open for almost a year. You are welcome to make a new PR when you have tests done.

PR #205, opened 2016-04-27
Add ker0x/CakeGCM plugin

ker0x:
https://github.com/ker0x/CakeGCM

This plugin allow you to send downstream message to an Android or iOS
device through Google Cloud Messaging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,9 +260,9 @@ Additional lists you might find useful:
 ## Notifications
 *Plugins for working with notification software.*
 
-- [Gcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream messages to an Android or iOS device through Google Cloud Messaging.
+- [ker0x/CakeGcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream messages to an Android or iOS device through Google Cloud Messaging.
 - [Notifier plugin](https://github.com/cakemanager/cakephp-notifier) - A plugin that makes creating and reading notifications easy.
-- [Push plugin](https://github.com/ker0x/cakephp-push) - A plugin to send push notifications through services like Firebase Cloud Messaging.
+- [ker0x/Push plugin](https://github.com/ker0x/cakephp-push) - A plugin to send push notifications through services like Firebase Cloud Messaging.
 
 ## ORM and Datamapping
 *Plugins that implement object-relational mapping or data-mapping techniques.*
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ Additional lists you might find useful:
 ## Notifications
 *Plugins for working with notification software.*
 
+- [ker0x/CakeGcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream message to an Android or iOS device through Google Cloud Messaging.
 - [Notifier plugin](https://github.com/cakemanager/cakephp-notifier) - A plugin that makes creating and reading notifications easy.
 
 ## ORM and Datamapping
--- a/README.md
+++ b/README.md
@@ -260,8 +260,9 @@ Additional lists you might find useful:
 ## Notifications
 *Plugins for working with notification software.*
 
-- [ker0x/CakeGcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream message to an Android or iOS device through Google Cloud Messaging.
+- [Gcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream messages to an Android or iOS device through Google Cloud Messaging.
 - [Notifier plugin](https://github.com/cakemanager/cakephp-notifier) - A plugin that makes creating and reading notifications easy.
+- [Push plugin](https://github.com/ker0x/cakephp-push) - A plugin to send push notifications through services like Firebase Cloud Messaging.
 
 ## ORM and Datamapping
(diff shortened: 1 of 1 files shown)


ADmad:
The functionality should be in a service class not a component. Having it component means one can't use it from a shell.

ayman-alkom:
Hi @ker0x 
I suggest you to rename the plugin to someting like **CakePush** because the `GCM` refered just to android servcie whereas apple service name is `APNS`

ADmad:
@ayman-alkom GCM is now available on ios too.

ker0x:
@ADmad 
Thanks for the advice. So I moved my code inside a single class Gcm.php in src/Lib as you can see here https://github.com/ker0x/CakeGCM/tree/develop

However, I've left the component for people who already using it in their applications, but it use the new class.

dereuromark:
@ker0x I think the idea would be to also at some point in the future allow the plugin to cover those other cases (maybe someone does a PR soon).

You can abandon the current repo and just make a new one with the suggested changes if you want.
Composer will then know that this is the new project.
All people using the current one will not face any problems.

ADmad:
@ker0x That's better, just I would suggest a more meaning folder like "Webservice" insted of "Lib". Have you checked that your updated component in `develop` branch actually works?  Cause you have overridden the [constructor](https://github.com/ker0x/CakeGCM/blob/develop/src/Controller/Component/GcmComponent.php#L29) and failed to called `parent::_construct()`, so i am quite skeptical :)  You should use the `initialize()` hook instead of constructor.

@dereuromark He can always rename/make a new repo later when he (or someone else) adds add support for APNS. Since the plugin only provides GCM support for now I don't see any problem with the plugin name.

bobmulder:
I really like the idea of supporting 'adapters'. For example, Amazon has also a service to send notifications, windows as well (for windows phones), and there must be more... So what about working with adapters? Take a look at http://flysystem.thephpleague.com/

ayman-alkom:
@ADmad no one said that the current name is a problem, it's just an idea aimed to make the plugin name better and bring some contributes for APNS.
after one week I will need to play with APNS and GCM, then I may contribute for adding APNS support.

@ker0x your plugin's name is fine if you dont plan to support other push services, in the future we may find another plugin with more general name then every body will target to it.

ADmad:
@ker0x Let us know when you have made a release with the changes you have on `develop` branch and we can merge this.

ker0x:
@ADmad New release available with the changes you've suggest :) https://github.com/ker0x/CakeGCM/releases/tag/v2.2.0

dereuromark:
Your documentation is wrong.
You need to be careful with type casing etc

ker0x/cake-gcm != ker0x/cake_gcm

ADmad:
@ker0x Ping, can you please fix the package name in your docs as @dereuromark suggested.

dereuromark:
Also "cakephp/cakephp": "~3.0", should be a require dependency, not a require-dev.
Or at least subpackages must be require. Right now it would mean this would work standalone with any PHP application which is not true.

ADmad:
Having cakephp/cakephp as dependency would be nice but not a necessity. If someone is dumb enough to think it's a standard alone package it's his fault :).

@ker0x I noticed that the builds are failing. Using `"phpunit/phpunit": "<6.0"` in composer.json should get the builds running.

dereuromark:
@ADmad But regarding the supported minors and majors it is still good to have at least something in require. This prevents accidental pull of an incompatible version.

ADmad:
Yup nice to have for that reason.

ker0x:
Hi everybody, sorry for the too long fix/response time!

@ADmad I have fixed the package name in the documentation, and made CakePHP a required dependency as suggested by @dereuromark. The compilation was failing as it was trying to send a notification to my device on which I deleted the test app. Anyway I found a nasty way to fix it, but it still appears as failing on the main page of the repo, which is not the case (https://travis-ci.org/ker0x/CakeGCM).

I also allowed myself to add another plugin to the README that I created, [Push] (https://github.com/ker0x/cakephp-push), which does the same thing as CakeGCM but for Firebase Cloud Messaging and uses adapters to make it easily expandable and add new services.

README.md
```diff
@@ -260,7 +260,9 @@ Additional lists you might find useful:
 ## Notifications
 *Plugins for working with notification software.*
 
+- [Gcm plugin](https://github.com/ker0x/CakeGCM) - A plugin to send downstream messages to an Android or iOS device through Google Cloud Messaging.
```
dereuromark: Isn't the actual plugin name still `ker0x/CakeGcm`?

README.md
```diff
@@ -260,7 +260,9 @@ Additional lists you might find useful:
...
 - [Notifier plugin](https://github.com/cakemanager/cakephp-notifier) - A plugin that makes creating and reading notifications easy.
+- [Push plugin](https://github.com/ker0x/cakephp-push) - A plugin to send push notifications through services like Firebase Cloud Messaging.
```
dereuromark: Same invalid name, should be 'ker0x/Push' obviously.

ker0x: Fixed

README.md
ker0x: Fixed

dereuromark:
Not sure about the order - if we order by plugin name with or without the vendor prefix. I think we so far use the former, incl. prefix/

PR #207, opened 2016-05-09
Suggesting swivel to the list

jrbasso:


README.md
```diff
@@ -330,6 +331,9 @@ Focus is on the specific task.
 - [Rocketeer](https://github.com/Anahkiasen/rocketeer) - Deployment
 - [Travis CI](https://travis-ci.org/) - A continuous integration platform - de-facto standard for testing (GitHub) repos.
 
+## Feature Toggle
+- [Swivel](https://github.com/zumba/swivel-cake) - A library that allows to setup up to 10 different behaviors with an interface for tracking results.
```
dereuromark: Isnt that a Cake Plugin?

jrbasso: @dereuromark Yes, it is a cake plugin that wraps the functionality of the standalone library (swivel).

ADmad: @jrbasso I think @dereuromark's inquiry was due to the phrasing "A library that...".  So perhaps rephrase the description to mention it's a plugin which wraps thats library.

Also would be better if you could put it one of the existing categories instead of a new one. There are already too many :)

dereuromark: Also the name would be

    Swivel plugin

Then :)

jrbasso: @dereuromark @ADmad Does the new update sounds better?

dereuromark:
Does this really need the extra category?

dereuromark:
If we can get rid of the new category I think this could be merged then. @jrbasso

jrbasso:
@dereuromark Change made. Sorry, for some reason I didn't read the previous comment.

dereuromark:
I will fix the alphabetical order post merge.

PR #208, opened 2016-05-18
Added RandomPrimaryKey Behavior

zvaehn:
Behavior for automatic random digit primary key creation for your model instead of AI.

https://github.com/51seven/cakephp-random-primary-key-behavior


dereuromark:
Tests missing.
Whats the use case, what happens if there is a (quite likely) collision, existing key etc?

ADmad:
Most common use case i can think of is not showing sequential / guessable IDs to user. For that one can use uuids primary key or just use primary key obfuscation (we already have plugins listed for that).

So I don't see a common enough use case for this behavior/plugin to be included in awesome list.

zvaehn:
thanks for this advice. Did not notice that there is already a plugin for that :)

PR #213, opened 2016-06-09
CakePHP Vagrant Setup

cpierce:
Added https://github.com/cpierce/cakephp-vagrant-setup to environment setups. This allows users to spin up a vanilla Cake PHP 3.x environment and immediately start the development process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,7 @@ Additional lists you might find useful:
 
 - [Cakebox](https://github.com/alt3/cakebox) - A Vagrant development environment powered by the CakePHP 3.x Console.
 - [CakePHP.gitignore](https://github.com/github/gitignore/blob/master/CakePHP.gitignore) - The .gitignore file proposals.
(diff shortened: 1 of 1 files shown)


ionas:
I am using puphpet for this - https://gist.github.com/ionas/d45e44b9576fdc9f2f02c3c25dfa3501 - however I really like your approach because it is little magic - and more options (esp to learn from) = great.

p.s.: do you setup databases to be my_app:my_app:secret by default?

cpierce:
@ionas - i set up the database to be dev:root:'' by default.  Would that be the preferred?

jhli:
I'd try to set it to defaults that match composer create-project cakephp/app

cpierce:
this now matches the create project defaults and runs on ubuntu 16.04 LTS with PHP 7 and mysql 5.7

cpierce:
Ping!

PR #218, opened 2016-08-13
Adding a .varci.yaml

lorenzo:
This may not be necessary for this repo as the amount of change requests is usually low, but is very cool nevertheless 💃 

This should at least help automate some of the tedium of maintaining the list and communicating with contributors.


ADmad:
Let's give this a try :+1:

dereuromark:
Do we need some docs/readme update or some contribution file explaining what to do here?

jadb:
Woohoo!! 🚀 

@dereuromark docs can be found here for now: https://github.com/VarCI/support/blob/master/pages/documentation.md. Let me know if you had something else in mind and I will get it PR'd later today.

Thanks for giving it a shot!

dereuromark:
I mean for people making a PR for this repo. They would need the info in a texual way to know what are the requirements for the PR to pass :)

ADmad:
I think the messages set for `message` keys is what will be shown to users.

.varci.yml
```diff
@@ -0,0 +1,43 @@
...
+      The broken link is: {{ diff_links.broken.0 }}
+    when:
+      - action = "opened" or action = "reopened"
+      - count(diff_links.broken) > 0
+      - files has "README.md"
+
+  link_in_description:
+    name: "the URL matches the URL in the PR's body"
+    events: [ pull_request ]
+    comment: >
```
ADmad: I think this key should also be `message` like others.

jadb: Fixed on master. The `message` key was deprecated (but still supported). Updated all to `comment`.

PR #221, opened 2016-08-22
does no plural

inoas:


jadb:
Hmm, need to cater for this use case (fixing existing link) on VarCI's side.

inoas:
I like my new personal overlord.
/me pats @VarCI-bot then bows to it!

bravo-kernel:
@jadb nice bot. 

**HTH:** this is my first acquaintance with the bot and the `Automatic update by Var.CI` message kind of confused me and made me think the bot actually updated something instead of just commenting. Maybe something like `Automatic comment/response by Var.CI` would be more appropriate?

jadb:
@bravo-kernel you are 💯  right!

bravo-kernel:
Probably just about it being a nice bot 😂

PR #222, opened 2016-09-01
Add DataSeed plugin

andrej-griniuk:
https://github.com/andrej-griniuk/cakephp-data-seed

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,7 @@ Additional lists you might find useful:
 ## Migration
 *Plugins and resources around migration and upgrading.*
 
+- [DataSeed plugin](https://github.com/andrej-griniuk/cakephp-data-seed) - Bake DB seed files with data included.
 - [Execution order](https://github.com/dereuromark/executionorder) - A 3.x test app to display the execution order of files, methods and callbacks.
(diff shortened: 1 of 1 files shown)


jadb:
So is this a select with limit to create the data seed? If that's the only functionality, wouldn't this have been much better contributed back to the migrations plugin?

If so, like its counterpart (fixture generation from database records) should be part of the core IMO.

jadb:
That doesn't mean I don't like the plugin! To the contrary, it just means I find its so important that it should be part of core. 

So thanks @andrej-griniuk!

andrej-griniuk:
Thank @jadb! Yep, that's pretty much all it does.

https://github.com/cakephp/migrations/issues/191
I saw this discussion and thought that quite a bit more is expected from this kind of feature. However I do believe that there is plenty of users like myself who only need as much as I've implemented here. So if you guys think that's good enough I'm more than happy to make a pull request to the migrations plugin instead.

dereuromark:
Lets try the pr for Migrations. If that's works awesome :) everyone uses that already anyway.

andrej-griniuk:
https://github.com/cakephp/migrations/pull/260

PR #225, opened 2016-09-28
Add DatabaseLog

dereuromark:
https://github.com/dereuromark/CakePHP-DatabaseLog

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,7 @@ Additional lists you might find useful:
 *Plugins for auditing and logging.*
 
 - [AuditStash plugin](https://github.com/lorenzo/audit-stash) - Flexible and rock solid audit log tracking.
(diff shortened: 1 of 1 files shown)


jippi:
LGTM; though the README lacks a huge warning that it very quickly can go out of hand performance and size wise logging to a database. 

Having a (semi)automated way to rotate logs when number of rows/table size exceed $x or simply by date range (e.g. auto-delete rows older than $x days with a gc probability / shell to do it)

inoas:
This is great stuff.

dereuromark:
Added note in https://github.com/dereuromark/CakePHP-DatabaseLog/blob/master/README.md#log-rotation

dereuromark:
The next minor will also natively contain a few shell commands for garbage collect and alike.

PR #229, opened 2016-10-31
Wording changes per dereuromark

cpierce:
@dereuromark - these changes should be better ... I was making them when you merged in my PR from before.


dereuromark:
The name plugin must stay though, cakephp packages are named "plugin". the rest looks good.

PR #233, opened 2016-11-09
DomPDF plugin

ozee31:
https://github.com/DaoAndCo/cakephp-dompdf

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,7 +268,7 @@ Additional lists you might find useful:
 *Plugins and software for working with PDF files.*
 
 - [CakePdf plugin](https://github.com/FriendsOfCake/CakePdf) - A plugin around PDF generation.
-- [DomPDF](https://github.com/DaoAndCo/cakephp-dompdf) - Generate PDF with DomPDF
+- [DomPDF](https://github.com/DaoAndCo/cakephp-dompdf) - Generate PDF with DomPDF.
 
 ## Queue
 *Plugins for working with event and task queues.*
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ Additional lists you might find useful:
 *Plugins and software for working with PDF files.*
 
 - [CakePdf plugin](https://github.com/FriendsOfCake/CakePdf) - A plugin around PDF generation.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -268,6 +268,7 @@ Additional lists you might find useful:
 *Plugins and software for working with PDF files.*
 
 - [CakePdf plugin](https://github.com/FriendsOfCake/CakePdf) - A plugin around PDF generation.
+- [DomPDF](https://github.com/DaoAndCo/cakephp-dompdf) - Generate PDF with DomPDF
```
dereuromark: Full stop missing. `Dompdf` seems to be their wording.

dereuromark:
Tests seem missing here.

ADmad:
What does this plugin offer that `CakePdf` doesn't?

ozee31:
This plugin allows you to easily generate PDF with dompdf. It uses the view's system of CakePHP

ADmad:
> This plugin allows you to easily generate PDF with dompdf. It uses the view's system of CakePHP

So does [CakePdf](https://github.com/FriendsOfCake/cakepdf)

ozee31:
True, it is an alternative

ozee31:
I created helper for help to generate images, css, page break...

dereuromark:
Fair enough.

But you do realize that this is not necessary (and shouldnt be) as the normal helper calls can easily be intercepted so you don't have to modify any code and can continue using the normal helpers?
See http://www.dereuromark.de/2014/04/08/generating-pdfs-with-cakephp/#tips for example.

ADmad:
That helper methods are pretty specific to your needs and not suitable for generic use. They have hardcoded paths and class names.

ozee31:
OK, I'll try to improve my plugin or try your plugin

ADmad:
Closing this PR for now as I don't think it provides anything that CakePdf already doesn't. Since this is meant to be a curated list and not a plugin registry we don't want to list multiple plugins for the same task unless they differ significantly.

ozee31:
OK no problem

ADmad:
You can submit your plugin to be listed on plugins.cakephp.org

ozee31:
ok

PR #237, opened 2016-12-01
What's the RPG pattern?

davidyell:
Role playing search!


dereuromark:
Oh geeze :) nice find!

PR #239, opened 2016-12-23
Adding CakePHP Fixtures plugin

LubosRemplik:
Tiny plugin, but might be useful for somebody.

https://github.com/LubosRemplik/CakePHP-Fixtures


dereuromark:
> "name": "lubos/fixtures"

It is usually recommended to prefix the plugin with cakephp, this now looks like a generic fixtures package, but it seems to in fact be a specific CakePHP plugin.

     "name": "lubos/cakephp-fixtures"

etc

It should also come with some tests.

LubosRemplik:
Thanks @dereuromark for your notes.

The package name was updated to cakephp-fixtures, some unit tests added.

dereuromark:
Looks good to me.

PR #241, opened 2017-01-10
Add CakePHP Interval plugin

LubosRemplik:
Converts seconds to human readable string (string to seconds), uses business hours (1 week = 5 days, 1 day = 8 hours). See readme for more info.

Maybe `Interval` name is too abstract, let me know if you have better idea for name.

https://github.com/LubosRemplik/CakePHP-Interval


dereuromark:
I like it. :+1: 
At some point in the future this could even be part of Chronos etc.

PR #246, opened 2017-02-18
Adding IdeHelper

dereuromark:
https://github.com/dereuromark/cakephp-ide-helper

The header here reads "parsing and manipulation codebases", so I thought maybe it fits into that description?


dereuromark:
Is the category fine?

PR #251, opened 2017-03-21
Adding Comments plugin

Kareylo:
Just added a plugin i just finish to comment whatever model we want !

https://github.com/Kareylo/CakePHP-Comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,6 +230,7 @@ Additional lists you might find useful:
 - [Ajax plugin](https://github.com/dereuromark/cakephp-ajax) - A plugin to ease handling AJAX requests.
 - [CakeAdmin plugin](https://github.com/cakemanager/cakephp-cakeadmin) - A non-stable user management plugin with a built-in admin area.
 - [CakeMiddlewares](https://github.com/chrisShick/CakeMiddlewares) - A collection of Cakephp Middlewares. Requires CakePHP version 3.4+.
+- [Comments plugin](https://github.com/Kareylo/CakePHP-Comments) - A fully customizable Comments plugin. Requires CakePHP version 3.3.2+.
 - [CurrencyConverter plugin](https://github.com/AlessandroMinoccheri/cakephp-currency-converter) - A plugin to convert currency into another one.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -230,6 +230,7 @@ Additional lists you might find useful:
 - [Ajax plugin](https://github.com/dereuromark/cakephp-ajax) - A plugin to ease handling AJAX requests.
 - [CakeAdmin plugin](https://github.com/cakemanager/cakephp-cakeadmin) - A non-stable user management plugin with a built-in admin area.
 - [CakeMiddlewares](https://github.com/chrisShick/CakeMiddlewares) - A collection of Cakephp Middlewares. Requires CakePHP version 3.4+.
+- [Comments plugin](https://github.com/Kareylo/CakePHP-Comments) - A fully customizable Comments plugin. Requires CakePHP version 3.3.2+.
```
dereuromark: The versions can be removed IMO. thats part of the readme and composer constraint.

dereuromark:
A few tests here would be nice to avoid regression.

Kareylo:
I begin to work on the differents tests. Will push them in the next release of the plugin when it's done.

Kareylo:
After a hard fight against travis, tests passed !

ADmad:
- If  you remove the `id` field values from your fixtures your test suite will then run properly on postgres too. Postgres doesn't like autoincrement values being set before hand. 
- Readme is missing instructions for loading the plugin (it might seem obvious but newbies can trip on it)
- I would suggest removing instructions of setting `$this->helpers` in controller. It's preferred to load behavior through `AppView` instead.

Kareylo:
@ADmad 
* If I remove the `id` field, every test using the field fail. Any example on how to setup that ?
* Forget to add that, my bad.
* Totally agree with you, directly remove from the readme

ADmad:
> If I remove the id field, every test using the field fail. Any example on how to setup that ?

Check the core fixtures or any plugin of FriendsOfCake.

dereuromark:
For me removing the versions would make this mergeable.

PR #256, opened 2017-03-27
Add Captcha plugin

dereuromark:
https://github.com/dereuromark/cakephp-captcha


bravo-kernel:
LGTM but isn't it time to stop using that PSR-2R? It's so damn confusing and actually keeps me from using your plugins, the mindf** when switching between plugins is just too much for me. No offense, just being honest here.

dereuromark:
It should rather be time to use it, it is far more superiour, and has tons of more and better standards on top.

PR #258, opened 2017-03-30
Add Crud to the REST/API section

bravo-kernel:
Since this might not be evident to (new) users, also promote JSON API a bit more.


bravo-kernel:
It's only 1 suggestion @VarCI-bot, the other two are whitespace cleaning.

dereuromark:
It is already in CRUD section, please move it here then if that is the better location.

bravo-kernel:
I realize it is in the `skeleton` section already but I feel it deserves to be in two places as it is both a skeleton plugin and an API plugin (IMO). Could an exception for a duplicate entry be tolerated here to better aid the end-user?

dereuromark:
I guess that is reasonable.

bravo-kernel:
Thank you kindly, anybody want to merge this?

PR #262, opened 2017-05-04
Update README.md

kushalpogul:
- [File Functions plugin](https://github.com/kushalpogul/file-functions) - A little Plugin to upload, download, delete file from FTP Server. This plugin also allows various useful file manipulation operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,7 @@ Additional lists you might find useful:
 ## Files
 *Plugins for file manipulation.*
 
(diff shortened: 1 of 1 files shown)


kushalpogul:
Please check now

ADmad:
File management functions don't belong in a component. They should be in a service class. 

There are already well known and well tested libraries like [flysystem](http://flysystem.thephpleague.com) for file management which provide adapter for FTP besides other file systems.

Thank you for your submission but it can't be featured on this list for reasons stated above.

PR #264, opened 2017-05-12
Replace gourmet/knp-menu with icings/menu.

ndm2:
[**`gourmet/knp-menu`**](https://github.com/gourmet/knp-menu) doesn't seem to be maintained anymore, just like Gourmet in general.

Given the current state, I'd like to propose replacing `gourmet/knp-menu` with [**`icings/menu`**](https://github.com/icings/menu), which is based on KnpMenu too, and ships with a string template renderer, and URL as well as route based matching.

I've used it for internal projects for quite some time now, but had to make some larger changes for the public release, so it's currently in beta. Not sure what the rules regarding stability are, but I thought throwing the idea in the room before the first stable won't hurt.

ping @phillaf

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,7 @@ Additional lists you might find useful:
 ## Navigation
 *Tools for building navigation structures.*
 
-- [KnpMenu plugin](https://github.com/gourmet/knp-menu) - A menu plugin based on the [Knp Menu Library](https://github.com/KnpLabs/KnpMenu).
+- [Icings/Menu plugin](https://github.com/icings/menu) - A [KnpMenu](https://github.com/KnpLabs/KnpMenu) seasoned menu plugin for CakePHP.
 
 ## NoSQL
 *Plugins for working with "NoSQL" backends.*


ADmad:
You have my vote for replacing `gourmet/knp-menu`. As you mentioned gourmet's plugin's aren't being maintained. Given that your plugin has 100% code coverage the beta status of the plugin is not an issue.

One thing I noticed in the docs [example](https://github.com/icings/menu/blob/master/docs/index.md#adding-menu-items) is it says for `['controller' => 'Articles', 'action' => 'index']` the URL generated `/articles/index`, is that correct?  As per Cake's default routes it should be `/articles`.

ndm2:
Not quite correct, no, intially this was all using this "neutral" `'controller' => 'Controller', 'action' => 'action'` config, and I just didn't update that properly in some places. Fixed now.

Phillaf:
Just here to second everything you guys said. Gourmet's plugin have [never been working correctly] (https://github.com/gourmet/knp-menu/issues/5) in the first place.

PR #268, opened 2017-06-26
updated

mirko-pagliai:
`cakephp-mysql-backup` has become `cakephp-database-backup`, to work with other database engines supported by cakephp (for now, MySql and Sqlite)


dereuromark:
https://github.com/mirko-pagliai/cakephp-database-backup

PR #270, opened 2017-07-06
Create cakephp3-login

Lakshmi-carmatec:
Create cakephp3-login


dakota:
I'm not sure what this is for, or how it relates to the Awesome CakePHP list, but it definately looks like it's in the wrong place...

PR #272, opened 2017-07-18
MenuLink plugin suggestion added

rrd108:
https://github.com/rrd108/cakephp-menulink


rrd108:
https://github.com/rrd108/cakephp-menulink

dereuromark:
This one little helper seems rather incomplete to me.
It only supports very basic controller action pairs, what about plugin or prefixes?
What about passed params, or customization?
Not sure this little code snippet here warrants a plugin listing.

ADmad:
Uhm... So this plugin is basically 2 lines of code. IMHO it doesn't provide enough value to warrant being on this list.

rrd108:
Thanks for the feedback, I will add them in the next version. :)

PR #274, opened 2017-07-22
Added simple environment manager

matscode:
Even though it does not have up to 10 stars yet as stated in CONTRIBUTION guide. I see it to be useful and easy to use, which why i wrote it. 

https://github.com/matscode/cakephp-simple-environment-manager

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,7 @@ Additional lists you might find useful:
 
 - [Environments plugin](https://github.com/josegonzalez/cakephp-environments) - Plugin to handle environments.
 - [Gourmet/Aroma plugin](https://github.com/gourmet/aroma) - Database based configuration.
+- [Simple Environment Manager](https://github.com/matscode/cakephp-simple-environment-manager) - Load right config file at the right Environment with ease and few setup.
 - [Settings plugin](https://github.com/cakemanager/cakephp-settings) - A plugin to manage your settings via your database.
 - [Setup plugin](https://github.com/dereuromark/cakephp-setup) - Plugin to handle very basic environments.
 


matscode:
https://github.com/matscode/cakephp-simple-environment-manager

dereuromark:
I dont see composer support etc.

dereuromark:
PS: Given the fact that https://github.com/cakephp/app/pull/526 and the PR before allows for env stuff natively, this here does not seem to offer much on top, does it?

ADmad:
As @dereuromark said since the 3.5 app skeleton has inbuilt support for .env files (which can be easily replicated in any existing app too), promoting another way for env based config seem unnecessary. Closing that reason.

matscode:
If you check the date of creation.. it was before the release of cakePHP V3.5.x.. moreso, not everybody gets the hang of .env files... **this makes use of internal cakePHP config override ability** lets see some pros and cons of this convention and using .env file below..

### `Simple Environment Manager`
###### Pros
- Helps manage environment based configuration
- Perfectly fits when working with team/openSource projects
- easy to update production config locally
- It automatically switch to use production config if `environment` is set to  `local` in `bootstrap.php`
- it does not require `.htaccess` to discretely protect config files in production as it uses `.php` based configuration
- it is really simple to implement in any environment
###### Cons
- requires over writing cakePHP default `bootstrap.php` file with custom
- it requires `.gitignore` when working with teamBased/OpenSource projects

### `.ENV`
###### Pros
- helps manage environment configuration
- Perfectly fits when working with team/openSource projects
###### Cons
- not every developer gets the hang of it
- remote log on is required to update production config
- requires  `.htaccess` file configuration to disable public visibility (else, it poses security issues and what if `.htaccess` gets deleted)
- it requires `.gitignore` when working with teamBased/OpenSource projects

With this, i believe i should i have convinced you a bit and if theres more i have not stated in the pros and cons. I dont mind learning by been added.

**PS**: will do an update to make it usable in cakePHP v3.5.x

ADmad:
Even if someone doesn't want to use .env files environment based config is as easy as loading an extra file based on some criteria [example](https://github.com/cakephp/app/blob/master/config/bootstrap.php#L83) for which is already shown in current app skeleton.

Or one can simply use a one liner like:
```
Configure::load('app_' . env(APP_ENV), 'default');
```

matscode:
Someone will have to manually comment out that line each time theres wanting to be a push to production... Moreso imagine i created more configuration files e.g `payment_api.php`, `twilio.php` etc.. and all this files uses a different environments (production, local, staging/development) based configuration... in simple environment manager all you have to do is add your default and override configuration file name to the configList ...

matscode:
and even if you forget to change your environment from `local` to `production` in production simple environment manager does some check and set the environment to `production` which will load production overrides instead of local overrides...

matscode:
i hope i am not taking much of your time @ADmad ...

matscode:
Should I go ahead make a new pull request? @ADmad @dereuromark

bravo-kernel:
Even though I can see you have invested :heart: in your solution I must say that I see no added value in your solution either mr. @matscode. 

FWIW I have been using `.env` a long time before 3.5 and thus far this has catered for all use cases you described (even with a static `.env` per environment). Besides that, I think it's very unwise to steer users away from what is core-supported and basically the de facto standard for environment based configurations. 

Might not be the feedback you want to hear but alas.

matscode:
:-D Now i see from your perspective. Nice meeting you guys... @bravo-kernel

PR #276, opened 2017-07-31
Crabstudio is now agiletechvn

anhtuank7c:
Change org link


ADmad:
Too bad composer/packagist still doesn't have a decent strategy to handle package renaming.
Related https://github.com/composer/packagist/issues/496.

PR #282, opened 2017-10-30
Remove CakeDC serch plugin

liviascapin:
Since it does not support CakePHP 3.x https://github.com/cakedc/search#requirements


dereuromark:
There is a 3.0 branch. But given the fact that FOR one is recommended as we facto standard we could indeed remove this one now.

ADmad:
There has never been a stable release of it's 3.0 branch nor has it had a commit in over a year, so it's pretty reasonable to remove it.

PR #284, opened 2017-12-18
Added a TokenVerify Plugin

mosaxiv:
This package provides an Easily issue tokens that can be used for mail authentication.

https://github.com/mosaxiv/cakephp-token-verify


ADmad:
You should clarity in your readme that the plugin generates JWT :slightly_smiling_face:.

The fact that your a using value of `modified` field to generate signing key concerns me a bit. It means if the record get's edited after JWT is generated the token will never validate.

mosaxiv:
Thanks for review.   Fixed. 😄 

By using modified field, JWT can be used as one-time tokens.  
JWT should be discarded when the table is updated.  
e.g. password reset, email activation, account confirmation.

ADmad:
You should add a clear cautionary note in the readme that the generated JWT will become invalid when user record is updated and then `modified` is updated. Also add an e.g showing how to change the default 10min expiry for the JWT.

mosaxiv:
add usage
https://github.com/mosaxiv/cakephp-token-verify#usage

PR #290, opened 2018-02-26
Update CakeDC Users

ajibarra:
* Update CakeDC Users plugin: add correct namespace as other plugins does.


ajibarra:
Thanks @dereuromark

PR #295, opened 2018-03-09
SweetAlertHelper plugin

falco442:
Added my plugin [sweet-alert-cakephp-3](https://github.com/falco442/sweet-alert-cakephp-3); it allows to use HtmlHelper::link and FormHelper::link with sweet alert 2, a beautiful alert system, alternative to javascript alert

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,7 @@ Additional lists you might find useful:
 - [Utils plugin](https://github.com/cakemanager/cakephp-utils) - Containing useful components (Authorizer, Menu) and behaviors (WhoDidIt, Uploadable, Metas, Stateable).
 - [Wrench plugin](https://github.com/HavokInspiration/wrench) - Maintenance Mode plugin. Easily extensible and customizable.
 - [Yaml plugin](https://github.com/guemidiborhane/Cake-Yaml) - For using YAML config files instead of PHP arrays.
-- [sweet-alert-cakephp-3](https://github.com/falco442/sweet-alert-cakephp-3) - To use `HtmlHelper::link` and `FormHelper::postLink` with confirmation dialog done with [Sweet Alert 2](https://sweetalert2.github.io/)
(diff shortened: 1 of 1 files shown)


dereuromark:
You have several issues here:

- composer name does not follow recommendations: cakephp-3- is now a good name here, as the major version should not be reflected - instead `cakephp-sweet-alert-helper` etc would work.
- You dont have any tests
- invalid name, see https://github.com/FriendsOfCake/awesome-cakephp/blob/master/CONTRIBUTING.md

dereuromark:
Also
- https://github.com/falco442/sweet-alert-cakephp/blob/master/src/Controller/AppController.php not needed

You dont seem to use any CS ( https://github.com/falco442/sweet-alert-cakephp/blob/master/src/View/Helper/FormHelper.php )
If you want to use [PSR2R](https://github.com/php-fig-rectified/psr2r-sniffer) (which it seems like), then you must include the sniffer and run it (also in travis). See e.g. my ide-helper plugin on how this is done.
Otherwise you must use sth like the cakephp default one.

falco442:
Thank you! I've never ever used a codesniffer, I installed the CakePHP one. Of course, i let it sniff the code.
I also changed the name of the plugin, removed the AppController that was unnecessary.

But, honestly, I didn't understand what does mean

> Name plugins as ExactPluginName plugin

Can you explain me?

(I still have to change the name of the plugin)

dereuromark:
See the syntax you have in your diff and what the document states?
[sweet-alert-cakephp-3] => [SweetAlertHelper plugin] given the fact that you renamed it to your-vendor-name/cakephp-sweet-alert-helper and the namespace of it is SweetAlertHelper then.

dereuromark:
Yes, but as i mentioned before it makes sense to remove the version number everywhere, especially the github package name.

falco442:
On github, the package name is sweet-alert-cakephp, but packagist doesn't allow me to change the package name; do I have to delete and re-submit the package name to packagist with the new name?

dereuromark:
Your composer require falco442/cakephp-3-sweet-alert-helper is still wrong on the readme afaik
And you dont have any tests yet it seems (only the illusion of it :smile: )

PR #303, opened 2018-06-05
Add filesystem plugin

josbeir:
Added "yet another" filesystem plugin :-)

=> https://github.com/josbeir/cakephp-filesystem

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,10 +165,10 @@ Additional lists you might find useful:
 *Plugins for file manipulation.*
 
 - [FileStorage plugin](https://github.com/burzum/cakephp-file-storage) - Abstract file storage and upload plugin.
-- [Filesystem plugin](https://github.com/josbeir/cakephp-filesystem) - Abstract [Flysystem](http://flysystem.thephpleague.com/) + file entity based upload plugin.
 - [FlyPie plugin](https://github.com/WyriHaximus/FlyPie) - Abstract filesystem access using Flysystem.
 - [FriendsOfCake/Upload plugin](https://github.com/FriendsOfCake/cakephp-upload) - A customisable plugin that uses [Flysystem](http://flysystem.thephpleague.com/) to write to multiple backends (Dropbox, FTP, S3, Local, etc.).
 - [Image plugin](https://github.com/josbeir/image) - Image behavior that works much like Cake's built in TranslateBehavior.
+- [Josbeir/Filesystem plugin](https://github.com/josbeir/cakephp-filesystem) - Abstract [Flysystem](http://flysystem.thephpleague.com/) + file entity based upload plugin.
 - [Proffer plugin](https://github.com/davidyell/CakePHP3-Proffer) - A customisable upload plugin with thumbnail generation.
 - [Upload plugin](https://github.com/Xety/Cake3-Upload) - A little plugin to upload file.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,6 +165,7 @@ Additional lists you might find useful:
 *Plugins for file manipulation.*
 
 - [FileStorage plugin](https://github.com/burzum/cakephp-file-storage) - Abstract file storage and upload plugin.
+- [Filesystem plugin](https://github.com/josbeir/cakephp-filesystem) - Abstract [Flysystem](http://flysystem.thephpleague.com/) + file entity based upload plugin.
```
dereuromark: Isnt the name technically `Josbeir/Filesystem` ?

josbeir: Technically yes but its a bit weird in representation? Are you proposing something? ;-)

dereuromark: Well, that's the name and thus the standard display name we are using to not confuse people.

josbeir: Fixed !

README.md
```diff
@@ -165,7 +165,7 @@ Additional lists you might find useful:
 *Plugins for file manipulation.*
 
 - [FileStorage plugin](https://github.com/burzum/cakephp-file-storage) - Abstract file storage and upload plugin.
-- [Filesystem plugin](https://github.com/josbeir/cakephp-filesystem) - Abstract [Flysystem](http://flysystem.thephpleague.com/) + file entity based upload plugin.
+- [Josbeir/Filesystem plugin](https://github.com/josbeir/cakephp-filesystem) - Abstract [Flysystem](http://flysystem.thephpleague.com/) + file entity based upload plugin.
```
dereuromark: Alph. order please.

josbeir: Fixed! Did a rebase, you're too fast :-)

josbeir:
Is there an 'evalution' period on these pr's ? :-)

dereuromark:
I think people were just hesitant because there are so many file handling ones already in the list.
Are there any key differences/improvements here?

I do like the docs verbosity :)

josbeir:
I can't speak of the other plugins but the 'cool' thing here is that instead of working with files directly the requirement is that we work with a file entity interface and a simple file entity class (nothing related to the ORM). There is no database requirement by default for storing and manipulating files.

But the beauty of this is that storing entities in the db can be easily done by just swapping the entity class configuration ``entityClass`` with an ORM one, this entity then needs to implement the `FileEntityInterface` interface and add a required method...

And of course we have events, php 7.1 strictness, working with Zend PSR streams, paths and $_FILES..., filename formatters, ...

dereuromark:
Well, I dont think it hurts to add if it is well documented and tested.
It does sound like a valuable addition.

josbeir:
I would think it passes those two requirements :-)

ADmad:
Why does the readme still say "(WIP)"? :)

josbeir:
@ADmad yeah, because it is still WIP, it is not battle tested in production as i'm still working on the project where it is made for. But it works pretty well :-)

I'll remove it because it will probably confuse people...

ADmad:
It can deter potential users, so better to remove. Also better to use `0.x` versioning until you are satisfied with your product :)

ADmad:
@dereuromark Good to merge?

dereuromark:
Where did the plugin load part go?
And composer require josbeir/cakephp-filesystem:dev-master should be only `composer require josbeir/cakephp-filesystem` when you have tags.

josbeir:
Correct, upated.

There is currently no need to load the plugin as there are no components, behaviors, routes, ...

PR #304, opened 2018-06-05
Add Rest plugin

narendravaghela:
https://github.com/sprintcube/cakephp-rest


davidyell:
I can't see anything in this plugin which cannot be achieved in the core already.

narendravaghela:
Just updated the repository URL - https://github.com/sprintcube/cakephp-rest

dereuromark:
@ADmad What shall we do here?

ADmad:
- Seeing things like `RestController` is quite off putting to me. The lone `beforeRender()` in the controller could go in a component instead. 
- The use of .ctp files for returning JSON in spite of using a custom JsonView seems quite unnecessary.

I am not comfortable promoting such practices.

dereuromark:
All right, let's close this for a revised and possible future contribution for 4.x then.

PR #305, opened 2018-07-11
datalist plugin added

rrd108:
https://github.com/rrd108/cakephp-datalist

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,7 @@ Additional lists you might find useful:
 - [Chocolate plugin](https://github.com/commercial-hippie/chocolate) - Front-End framework FormHelper extensions.
 - [CommonMark plugin](https://github.com/gourmet/common-mark) - Adds [CommonMark](http://commonmark.org) (markdown) support to models and views.
 - [CsvView plugin](https://github.com/FriendsOfCake/cakephp-csvview) - A view class to easily generate CSV.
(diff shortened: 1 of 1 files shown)


ADmad:
Instead of `Plugin::load('Datalist'); ` the readme should suggest using the plugin shell to load the plugin so that apps using cake 3.6+ use the new method of plugin loading.

dereuromark:
Even though

    "cakephp/cakephp": "^3.5"

so technically this is the correct command to provide BC for 3.5.

ADmad:
This asserting doesn't seem right.

https://github.com/rrd108/cakephp-datalist/blob/0c0ca3c2e8e88c6ec60bb2f6001328b32f7ba28e/tests/TestCase/Model/Behavior/DatalistBehaviorTest.php#L74

Having `'issuer_id'` in data implies a belongs to relation in which case the association property name would be `company` and a single entity not an array of entities?

ADmad:
It would be better if you wrote tests using models which use fixtures.

rrd108:
I have fixed the following
- travis build image added to README
- plugin load example uses console in the README
- test cases refactored to use table objects
- new test cases added
- fixed a bug when there were more datalists and one of them got a selected an the other a new value

I think what I did not fixed is to add CO badge to the README suggested by @dereuromark as I would need some hint how to do that :)

rrd108:
Should I do anything else?

PR #307, opened 2018-07-19
Propose to add liqueurdetoile/cakephp-orm-json plugin

liqueurdetoile:
Hi,

I was working with an EAV data model and was needing some JSON fields to avoid value rows count exponential growth (I know, I know, that's evil for some people 😈).

Anyway, I began to write some inner stuff to handle it and, lastly, think about making it as a plugin : [https://github.com/liqueurdetoile/cakephp-orm-json](https://github.com/liqueurdetoile/cakephp-orm-json)

So here I am 😊

It only supports Mysql databases as v1.0.0 but I think it can easily be improved (see todo section of the readme)

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ Additional lists you might find useful:
 ## ORM and Datamapping
 *Plugins that implement object-relational mapping or data-mapping techniques.*
 
+- [Access JSON fields inner data from ORM](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM
 - [ADmad/Sequence plugin](https://github.com/ADmad/cakephp-sequence) - Behavior for maintaining ordered list of records.
 - [Duplicatable plugin](https://github.com/riesenia/cakephp-duplicatable) - Behavior for duplicating entities including related data.
 - [JeremyHarris/LazyLoad plugin](https://github.com/jeremyharris/cakephp-lazyload) - An association lazy loader for entities.
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ Additional lists you might find useful:
 - [ADmad/Sequence plugin](https://github.com/ADmad/cakephp-sequence) - Behavior for maintaining ordered list of records.
 - [Duplicatable plugin](https://github.com/riesenia/cakephp-duplicatable) - Behavior for duplicating entities including related data.
 - [JeremyHarris/LazyLoad plugin](https://github.com/jeremyharris/cakephp-lazyload) - An association lazy loader for entities.
-- [LiqueurdeToile/OrmJson plugin](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM.
+- [Lqdt/OrmJson plugin](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM.
 - [Money plugin](https://github.com/gourmet/money) - Money data type for CakePHP entities using [sebastianbergmann/money](https://github.com/sebastianbergmann/money).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -278,6 +278,7 @@ Additional lists you might find useful:
 ## ORM and Datamapping
 *Plugins that implement object-relational mapping or data-mapping techniques.*
 
+- [Access JSON fields inner data from ORM](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM
```
dereuromark: Please see the contribution guidelines. Several violations here.
One being the name of the plugin, more like [Lqdt/Coj plugin] probably
I do not like this nonspeaking name however, as it does not contain the intend inside.
VendorName/Purpose usually is better, in your case VendorName/OrmJson for example.

dereuromark:
If you also add the full stop at the end as per specs I am happy :)

README.md
```diff
@@ -281,6 +281,7 @@ Additional lists you might find useful:
 - [ADmad/Sequence plugin](https://github.com/ADmad/cakephp-sequence) - Behavior for maintaining ordered list of records.
 - [Duplicatable plugin](https://github.com/riesenia/cakephp-duplicatable) - Behavior for duplicating entities including related data.
 - [JeremyHarris/LazyLoad plugin](https://github.com/jeremyharris/cakephp-lazyload) - An association lazy loader for entities.
+- [LiqueurdeToile/OrmJson](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM
```
dereuromark: word plugin missing as per specs.

dereuromark:
As I noted above: You can't just change the name without adjusting the code itself.
E.g. https://github.com/liqueurdetoile/cakephp-orm-json/blob/master/src/Plugin.php#L2

dereuromark:
A question towards to the content:
A custom (Json)Type object didnt help you here? For me this was always useful when working with storing array or json data in the database.

liqueurdetoile:
I thought about custom type way but, if I understand well, you can only act on value ?
The plugin turns field name from 'pathInJson@field' into 'field->"$.path"' in queries.
PDO statements are also working very bad with JSON_EXTRACT data comparison so I needed to rewrote a full query for some data types like boolean, null, array or float. At least strings are still escaped with prepared statement.
I've found that extending Query object was the best approach but maybe I was wrong ?

README.md
```diff
@@ -281,6 +281,7 @@ Additional lists you might find useful:
 - [ADmad/Sequence plugin](https://github.com/ADmad/cakephp-sequence) - Behavior for maintaining ordered list of records.
 - [Duplicatable plugin](https://github.com/riesenia/cakephp-duplicatable) - Behavior for duplicating entities including related data.
 - [JeremyHarris/LazyLoad plugin](https://github.com/jeremyharris/cakephp-lazyload) - An association lazy loader for entities.
+- [LiqueurdeToile/OrmJson plugin](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM.
```
dereuromark: Name is now wrong.

arturmamedov:
Off topic - @liqueurdetoile why it is evil for some peaple, i think it a great approach for extra and optional fields in database, nor?

liqueurdetoile:
Wow. Some people have also dedicated whole blogs and books to the question !

To make it short. RDBMS like Mysql expects data to be described through a schema to give their best performance because they are working in a grid of rows and columns. Each row is a record and each column is a field value. The goal of RDBMS is to maintain data integrity through ACID operations. Using JSON fields is like a dark area for a pure RDBMS. It is not designed to be indexed and can lead to really poor performances in case of abuse. That's the technical evil side of non schemed data in RDBMS.

On the opposite, pure document database systems like MongoDB are really versatile. It only relies on programmer knowledge to fill in right data in documents.

But, I agree with you. Between the pros and cons of pure ruled world vs pure free world, we can be in need of both. It all relies on the objectives.

Most important : avoid using JSON fields with RDBMS because of too lazy to describe data. That's the human evil side of non schemed data in RDBMS.

PR #308, opened 2018-09-23
Add calendar plugin

dereuromark:
Adds https://github.com/dereuromark/cakephp-calendar

See 
- https://sandbox.dereuromark.de/sandbox/calendar for calendar example
- https://github.com/dereuromark/cakephp-calendar/tree/master/docs#icalendar-files-icalics for ics generating


ADmad:
`Plugin::load('Calendar');` should be removed from readme. It will throw warning in 3.6 :slightly_smiling_face:.

PR #309, opened 2018-09-30
Update README.md

kicaj:
Add another Slug plugin


ADmad:
How is this plugin different than the other two already on the list or what feature does it provide which is not available in the other two?

ADmad:
Checking your README I don't see anything which cannot be already achieved by using `muffin/slug` for example. So I am sorry but we can't include your plugin. This is a curated list not a plugin registry.

P. S.  You should also add tests for your plugin so that users have some confidence in the quality of code.

PR #311, opened 2018-10-01
Add Heartbeat Plugin

liviascapin:
Adds https://github.com/orca-services/cakephp-heartbeat

The heartbeat plugin is a CakePHP plugin providing an application heartbeat status page. It comes with a few sensors, which are configurable, and can easily be extended to e.g. monitor an API. 

We use it in combination with wormly (https://www.wormly.com/), to keep track of our applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ Additional lists you might find useful:
 - [Feedback plugin](https://github.com/dereuromark/cakephp-feedback) - Allow visitors to send quick and easy feedback incl. a screenshot via sidebar form.
 - [Flash plugin](https://github.com/dereuromark/cakephp-flash) - More powerful flash messages for your application.
 - [Hashid plugin](https://github.com/dereuromark/cakephp-hashid) - Allows to use hashids to not expose the database ids to the user.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -248,6 +248,7 @@ Additional lists you might find useful:
 - [Feedback plugin](https://github.com/dereuromark/cakephp-feedback) - Allow visitors to send quick and easy feedback incl. a screenshot via sidebar form.
 - [Flash plugin](https://github.com/dereuromark/cakephp-flash) - More powerful flash messages for your application.
 - [Hashid plugin](https://github.com/dereuromark/cakephp-hashid) - Allows to use hashids to not expose the database ids to the user.
+- [Heartbeat plugin](https://github.com/orca-services/cakephp-heartbeat/) - Monitor the hearbeat of your application (e.g. whether the database is available and up-to-date).
```
ravage84: How about linking the 3.x relevant branch directly? 
https://github.com/orca-services/cakephp-heartbeat/tree/cakephp-3.x

dereuromark: We dont do that. as this would mean we always have to modify this all the time.

PR #316, opened 2019-01-02
Add CakeImpersonate Plugin

challgren:
https://github.com/jomweb/CakeImpersonate

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,7 +243,7 @@ Additional lists you might find useful:
 - [CakeAdmin plugin](https://github.com/cakemanager/cakephp-cakeadmin) - A non-stable user management plugin with a built-in admin area.
 - [CakeDC/Enum](https://github.com/CakeDC/enum) - A plugin to add enumeration list support to your app.
 - [CakeDto plugin](https://github.com/dereuromark/cakephp-dto) - Quickly generate useful data transfer objects for your app (mutable/immutable), replacing messy arrays and leveraging your IDE through typehinting and autocomplete.
-- [CakeImpersonate plugin](https://github.com/jom/CakeImpersonate) - A component that stores the current authentication session and creates new session for impersonating Users. User can revert back to original authentication sessions without the need to re-login.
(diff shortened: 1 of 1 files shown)


dereuromark:
The package doesnt have any tests yet. That blocks the listing.
Any reason why it doesnt use the recommended MIT license?

Apart from that looks like a useful thing for debugging or helping people to do a task.
I usually hacked this into each app manually where I needed it.

challgren:
I've done 3 PRs to the repo the last 12 hours so it will be 100% covered and 100% tested using the setup from dereuromark/cakephp-tools. Thank you for your setup BTW. It has 55 downloads on Packagist and the repo owner seems to be receptive of PRs. I'm not 100% why @ akutaktau choose the license.

ADmad:
For me the plugin is too simplistic in it's current state to warrant adding to the awesome list. All it does is switch arrays in session.

dereuromark:
I agree that the functionality could be more complete.
Let's close until it can be revisited in the future.

challgren:
Ok the plugin has been greatly upgraded. Supports CI builds, uses correct methods for testing with 100% code coverage on Codecov. Does more work automatically such as using the same finder the AuthComponent uses along with using the UserModel from AuthComponent. The docs still need improvement but it's a lot better then when this PR was submitted.

challgren:
I will see if we can switch the license to be MIT.

ADmad:
- Why is the `updateConfig()` method called for both `Controller.initialize` and `Controller.startup` events?
- The `$name` property should be set in `Plugin` class.
- `AppController` and `webroot` directory are unnecessary.
- I don't see the point of the route set in `routes.php`.

challgren:
updateConfig() gets the UserModal and finder used in the `AuthComponent` to match the expected array for `setUser();` the events are called incase the user has updated the Auth config in their methods. _Reverse engineers how AuthComponent is setup and rebuilds the correct array_

The routes.php will be deleted in 3.0 which is under dev right now. It's going to take a more secure by default setup. And it will allow the SecurityComponent and CsfrComponent a chance to inspect the request to assess something risky.

Will update the recommendations in the 3.dev branch.

challgren:
License has been updated to MIT and 3.0.0-RC1 has been released.

challgren:
Version 3.0 is released and stable

PR #317, opened 2019-02-25
Adding LDAP, CakeSoap, and CakeSpreadsheet

darensipes:
- https://github.com/QueenCityCodeFactory/LDAP
- https://github.com/QueenCityCodeFactory/CakeSoap
- https://github.com/QueenCityCodeFactory/CakeSpreadsheet


dereuromark:
Are you sure the plugin names are as they are in code?

ADmad:
None of your plugins have tests.

darensipes:
@dereuromark Not sure what you are asking? They are named the same from what I can tell.

@ADmad Correct, we have no tests for these plugins currently. We could add them. Do we need the tests for them to be added to the list? Tests are something I could use help with, to be honest. I have minimal experience writing tests.

dereuromark:
https://github.com/QueenCityCodeFactory/LDAP/blob/master/composer.json#L33
is the correct plugin name. Please respect casing.
Also install docs should be complete etc.

dereuromark:
License files are missing! As per Guidelines.

ADmad:
Reasonable amount of time has passed and still no tests have been added to the linked plugins, hence closing the PR.

PR #318, opened 2019-04-19
Add LinkScanner

mirko-pagliai:
Adds https://github.com/mirko-pagliai/cakephp-link-scanner


ADmad:
Not a good idea suggesting deprecated method `Plugin::load()` in plugin meant for 3.7+.

mirko-pagliai:
> Not a good idea suggesting deprecated method `Plugin::load()` in plugin meant for 3.7+.

it's just a copy-paste in the README that refers to the cookbook. I can obviously delete the example and leave only the reference to the manual.

ADmad:
Best to suggest the CLI command `bin/cake plugin load LinkScanner`.

mirko-pagliai:
Fixed.

PR #319, opened 2019-04-23
Add CakePHP+Vuejs Demo App

ishanvyas22:
Adds https://github.com/ishanvyas22/cakephpvue-spa


dereuromark:
You could also add a free heroku live demo site link. For example.

ishanvyas22:
Sure, that is good suggestion. I will implement that in the future.
Thanks.

dereuromark:
See my sandbox (even though run on actual servers^^)

PR #322, opened 2019-05-11
Add HaveIBeenPwned plugin

FinlayDaG33k:


ADmad:
- The plugin's readme barely gives any idea about what it is for. Users' should shouldn't to google to figure out what's "HIBP API". Also it should have proper installation and usage instructions.
- The plugin doesn't have tests.
- Why are there empty templates when the controller returns JSON?
- Where can one find info about the service used in `HIBPComponent`? haveibeenpwned.com's API is pretty different https://haveibeenpwned.com/API/v2.

TLDR; In it's current state your plugins does not meet the standards we expect to be featured on this list.

ADmad:
The plugin isn't even installable using composer. Please read our [Contribution Guidles](https://github.com/FriendsOfCake/awesome-cakephp/blob/master/CONTRIBUTING.md)

There's lots of do here. Please re-open the PR after the various issues are addressed.

FinlayDaG33k:
That's odd? I'm pretty sure I did all of that? (might be that it's still on my old workstation and forgot to push it).

Will re-open the PR after the repo meets the QG.

PR #324, opened 2019-07-18
Update README.md

mirko-pagliai:
Adds https://github.com/mirko-pagliai/cakephp-stop-spam

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,7 @@ Additional lists you might find useful:
 - [Muffin/Throttle plugin](https://github.com/usemuffin/throttle) - A plugin for rate limiting (API) requests.
 - [Recaptcha plugin](https://github.com/agiletechvn/Recaptcha) - Simple, lightweight Google Recaptcha v2.
 - [Recaptcha Mailhide plugin](https://github.com/mirko-pagliai/cakephp-recaptcha-mailhide) - A plugin that allows you to hide email addresses using reCAPTCHA.
+- [StopSpam plugin](https://github.com/mirko-pagliai/cakephp-stop-spam) - A plugin that allows you to check if a username, email address or ip address has been reported as a spammer.
 
 ## SEO
 *Search Engine Optimization.*


README.md
```diff
@@ -344,6 +344,7 @@ Additional lists you might find useful:
 - [Muffin/Throttle plugin](https://github.com/usemuffin/throttle) - A plugin for rate limiting (API) requests.
 - [Recaptcha plugin](https://github.com/agiletechvn/Recaptcha) - Simple, lightweight Google Recaptcha v2.
 - [Recaptcha Mailhide plugin](https://github.com/mirko-pagliai/cakephp-recaptcha-mailhide) - A plugin that allows you to hide email addresses using reCAPTCHA.
+- [StopSpam plugin](https://github.com/mirko-pagliai/cakephp-stop-spam) - A plugin that allows you to check if a username, email address or ip address has been reported as a spammer.
```
dereuromark: ip => IP

dereuromark:
$this->Client->get('http://api.stopforumspam.org/api'

This does not yet work via https in 2019? That's a bit sad, also in terms of security.

mirko-pagliai:
> ```
> $this->Client->get('http://api.stopforumspam.org/api'
> ```
> 
> This does not yet work via https in 2019? That's a bit sad, also in terms of security.

Good point

mirko-pagliai:
Done. Now it uses `https`

dereuromark:
Looks good and useful to me.
I so far used my own "undisposable" lib and tooling, maybe I can sync it with this one in the future.

mirko-pagliai:
@dereuromark over time, I realized that all traditional methods (captcha and others) were often ineffective.
With the sites I manage, although I use reCaptcha, I continued to receive a lot of spam from the "contact me" form.
So I simply found it more useful to rely on a large database where users manually add spammers. So the plugin checks if a value has already been reported as a spammer and saves the values already verified in the cache (so as not to have to make continuous GET requests), while the app uses the session to understand if the user's ip control is already done.

If someone wants to contribute or improve, I'm happy. Bye.

dereuromark:
For bot fake systems you cannot rely on session though, they are auto busting this or using a distributed network. Even IP is sometimes not effective here though.
Often it is a combination of things that is really effective.

PR #325, opened 2019-08-19
Remove false statement about NewRelic plugin

Spriz:
I believe the NewRelic plugin does depend on the newrelic agent very much: https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Lib/NewRelic.php

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Additional lists you might find useful:
 ## APM
 *Plugins for Application Performance Monitoring.*
 
-- [NewRelic plugin](https://github.com/jippi/cakephp-newrelic/tree/cake3) - A complete plugin that enables full New Relic integration for a CakePHP application, including CLI naming, exceptions sending, custom timings, etc. It does not depend on New Relic agent.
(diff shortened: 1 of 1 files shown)


dereuromark:
I agree that the dependencies are not valid and that it requires certain classes to be there to function ( https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Traits/NewRelicTrait.php )
Composer should be fixed here.

Spriz:
@dereuromark to increase the confusion that `NewRelic\Lib\NewRelic` does actually live within the plugin ;P (https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Lib/NewRelic.php#L2)

It does however use methods from the PHP extension `newrelic` and does nothing in case that has not been loaded (https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Lib/NewRelic.php#L30-L36, https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Lib/NewRelic.php#L45-L47, https://github.com/jippi/cakephp-newrelic/blob/cake3/src/Lib/NewRelic.php#L65-L67 etc)

dereuromark:
OK, but this detail should then be in the README itself, that should suffice as info.

PR #326, opened 2019-10-07
added Fetchable behavior

segy:
https://github.com/riesenia/cakephp-fetchable

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,7 @@ Additional lists you might find useful:
 
 - [ADmad/Sequence plugin](https://github.com/ADmad/cakephp-sequence) - Behavior for maintaining ordered list of records.
 - [Duplicatable plugin](https://github.com/riesenia/cakephp-duplicatable) - Behavior for duplicating entities including related data.
+- [Fetchable plugin](https://github.com/riesenia/cakephp-fetchable) - Behavior for fetching entities from cache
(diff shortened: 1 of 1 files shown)


dereuromark:
Full stop is missing at the end.

As for your plugin:
- missing tests

segy:
thanks... good point. i'll update and let you know

segy:
Hi @ADmad - no problem... I don't force you to use it :) Just pointing out differences - you have `fetchEntity` method available (of course you can add additional one in addition to custom finder to your model, but this means extra writing) plus you have one extra check that even prevents cache engine to be hit. 

As I write in README, it's "Relevant for tables that contain moderate number of rows and are used commonly in many parts of application". I don't say, that this is a common case, but I used it a lot before, so I made it a public package so it can maybe serve someone else.

Btw @dereuromark issues you pointed out are fixed.

dereuromark:
OK
@ADmad shall we merge this?

ADmad:
@segy Since you are using return type hints you should specify the PHP version constraint in your composer.json https://github.com/riesenia/cakephp-fetchable/blob/master/src/Model/Behavior/FetchableBehavior.php#L89

@dereuromark I am fine with merging after the issue I pointed out if fixed. There will probably be others who find this useful :)

dereuromark:
I agree.

dereuromark:
ping @segy Can we proceed here?

segy:
hi. sorry for the delay. i added the php constraint.

PR #327, opened 2019-10-26
Adds inertia.js adapter

ishanvyas22:
Adds https://github.com/ishanvyas22/cakephp-inertiajs


dereuromark:
- Invalid name, please see guidelines.
- no tests

as for your naming: InertiaCake is confusing when the package name is ishanvyas22/inertia-cakephp maybe.
Most prefix it here with Cake instead of suffix: CakeInertia
But not a big deal.

dereuromark:
ping @segy Can we proceed here?

ishanvyas22:
Yes

dereuromark:
I dont see any of the first two critical points adressed, though.
Closing then for now. It requires a fresh PR with all the guidelines addressed.

PR #328, opened 2019-10-28
Add AssetMix plugin

ishanvyas22:
Adds https://github.com/ishanvyas22/asset-mix plugin

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,10 +81,10 @@ Additional lists you might find useful:
 *Tools for managing, compressing and minifying website assets.*
 
 - [AssetCompress plugin](https://github.com/markstory/asset_compress) - A complete asset manager for CakePHP.
+- [AssetMix plugin](https://github.com/ishanvyas22/asset-mix) - Provides integration with [Laravel Mix](https://laravel-mix.com) asset compilation.
 - [Assets plugin](https://github.com/mirko-pagliai/cakephp-assets) - Dynamic and "on the fly" asset files.
 - [Less plugin](https://github.com/elboletaire/less-cake-plugin) - Less parser plugin for CakePHP.
 - [MinifyHtml plugin](https://github.com/WyriHaximus/MinifyHtml) - Compress HTML output.
-- [Asset Mix](https://github.com/ishanvyas22/asset-mix) - Provides integration with your CakePHP application & [Laravel Mix](https://laravel-mix.com).
 
 ## Auditing / Logging
 *Plugins for auditing and logging.*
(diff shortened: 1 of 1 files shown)


dereuromark:
- Invalid name, please see guidelines.
- no tests

Otherwise I really like the idea of having such a tool in Cake now :) Nice one!

PS: echo $this->Mix->script('app');
You might want to clarify that this should either be done at the bottom of the page (before closing body tag), or one should be using the "block=>true" option to append to scripts until echoed at the end then in similar way.

dereuromark:
Also, your name is somewhat in conflict with the composer name, which also can create further confusion.

`"Mix\\": "src/"` vs `ishanvyas22/asset-mix`

Why not using AssetMix here for the namespace to avoid this?

dereuromark:
And as for your releases:
Why would you release readme only changes (semantic only) as major 0.2.0 version?
All 0.x are majors by definition. What you usually want is patches. But for semantic only usually that is also not needed. Maybe sth to watch out for for future releases.

dereuromark:
Also, shouldnt env() be used instead of directly reading from magic superglobal $_SERVER?

ADmad:
`config` directory isn't exactly the right place for functions file https://github.com/ishanvyas22/asset-mix/blob/master/config/functions.php

Also why are including it from bootstrap instead of  loading it through composer?

ishanvyas22:
@dereuromark, @ADmad thanks for your feedback.


> Invalid (plugin) name in overview, please see guidelines.

I've changed plugin name to `AssetMix` 

> PS: echo $this->Mix->script('app');
> You might want to clarify that this should either be done at the bottom of the page (before closing body tag), or one should be using the "block=>true" option to append to scripts until echoed at the end then in similar way.
> Only CSS should be outputted in the head of HTML pages.

Added `defer` attribute by default in all script tags to overcome this. Hope it's fine?

> Also, shouldnt env() be used instead of directly reading from magic superglobal $_SERVER?

Changed.

> config directory isn't exactly the right place for functions file https://github.com/ishanvyas22/asset-mix/blob/master/config/functions.php. It should be directly under src.

Nice suggestion, changed.

> This means it will generate incorrect URLs if the app is in a subfolder of doc root.

Don't know what you mean by that. Please elaborate.

Also, I've released this changes as a [minor release](https://github.com/ishanvyas22/asset-mix/releases/tag/0.3.0) since changing plugin name can break the system. @dereuromark  hope this is fine :)

For tests, I'm working on it but I really not good at testing yet. I would appreciate any help or contribution in this.

dereuromark:
You can look into the many plugins in the awesome list, they all have tests provided as showcase for you.

ishanvyas22:
Tests added, please review and merge.

dereuromark:
You can add batches to the readme top, see e.g. my plugins.
This helps to navigate to the test results as well as show that they are still passing.
Also, it can show coverage and packagist infos etc.

PR #329, opened 2019-11-13
Added the Jira Plugin.

fr3nch13:
Added the Jira Plugin I wrote under the Third Party APIs section.

https://github.com/fr3nch13/cakephp-jira


dereuromark:
Your skeleton looks a bit low on files, see other plugins.
Also seems to be missing tests and other requirements as per the howto of this repo.

You could probably use mocks to have tests without having to interact with a real Jira API.

fr3nch13:
I've updated the plugin to include tests and docs. Let me know if I'm missing anything else.

fr3nch13:
Please re-review. I've added more functionality, tests, CI checks and code coverage.

dereuromark:
Tmp should not be committed.

fr3nch13:
That’s where my coverage is getting placed, but touché, I’ll exclude it, and update the CI to create it.

PR #333, opened 2019-12-12
CakePHP Domain Manager added

pabloelcolombiano:
* I received some positive feedback from other users.
* My website N8Wedding.de, which uses multiple plugins, works entirely using the domain splited MVC. I tried all kind of cases: it works.
* I have written a few tests too.
* The package is very stable so I guess it is time to share.
* Thanks for reviewing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ Additional lists you might find useful:
 - [CurrencyConverter plugin](https://github.com/AlessandroMinoccheri/cakephp-currency-converter) - A plugin to convert currency into another one.
 - [Dashboard plugin](https://github.com/gourmet/dashboard) - Build beautiful dashboards for your cakes.
 - [DatabaseBackup plugin](https://github.com/mirko-pagliai/cakephp-database-backup) - A plugin to export, import and manage database backups.
+- [DomainManager plugin](https://github.com/pabloelcolombiano/cake-dm) - Refactor your monster MVC in smaller domain specific MVCs.
 - [Feedback plugin](https://github.com/dereuromark/cakephp-feedback) - Allow visitors to send quick and easy feedback incl. a screenshot via sidebar form.
 - [Flash plugin](https://github.com/dereuromark/cakephp-flash) - More powerful flash messages for your application.
 - [Hashid plugin](https://github.com/dereuromark/cakephp-hashid) - Allows to use hashids to not expose the database ids to the user.


ADmad:
From the plugin's readme.

> You would end up with a quantity of plugins added to the real plugins, and a deserted src folder.

Why is that a problem? Organizing your complete app using plugins only is perfectly reasonable.

Based on the example folder structure shown all you have done is reimplement the plugin structure within the app's `src` folder itself (albeit with the possibility of nested folders).

Your folder restructuring isn't domain driver design. It's just grouping of MVC classes within folders. This plugin is a proper example of DDD https://github.com/burzum/cakephp-service-layer.

ADmad:
> It is therefore forbidden to have two controllers with the same name in different domains.

That's a highly restrictive limitation killing all the benefit of using namespaces.

pabloelcolombiano:
@ADmad thanks a lot for the input. To your comments:

1. 

> Why is that a problem? Organizing your complete app using plugins only is perfectly reasonable.

It is not a problem to organize your app by putting all you domains in plugins, I agree. This is definitely one appropriate and elegant way to do it, no discussion about it.

In my problematic app with ~200 controllers and more or less as much tables, the refactoring effort and risks implied by reorganizing the code in plugins is very high. I would have some routing work, links to change, elements to rename, basically the whole app needs to be resorted.

But letting the refactoring time and risk aspect aside. In my app's strategy, plugins are customer features. One client loads a plugin, or not, and this way I have a clean separation of the client features. Maybe not optimal, but that is the way I went. I also have some home-made plugins like pdf rendering, email management, admin section and whatever would have been too big to fit in a component.

In the domain-plugins approach you suggest, I do not have a way to clearly distinguish between client feature plugins, home made plugins, or a "domain" plugin. I could name the domain-plugins DomainPluginFoo though, so they pop out from the rest. But I think it is a shame to leave "/src" empty and throw everything in the "/plugins" folder. But this applies to my case.

2. 

> Based on the example folder structure shown all you have done is reimplement the plugin structure within the app's src folder itself (albeit with the possibility of nested folders).

Yes, nothing more. It's pretty straighforward.

3.

> Your folder restructuring isn't domain driven design.

That is a topic. To put my view on that in short: service layer is a thing, structuring one's code in a readable maintainable way another. Both are aspects of DDD to my view.

I really like the cakephp-service-layer package, and gave it a successfull try in a past project.

My problem here is not about separating business logic from the framework structure, this can be done in multiple manners, which is out of the scope of the package. The aim of the package is to be able to clearly say to a new comer or to someone not aquainted with a given region of the code: "Here is the Domain Foo with its 10 controllers, tables, templates and so on, and don't care about the rest". Or to hear it from someone.

4.

> I agree that the wording "domain" or "layer" can be misleading. Any suggestion?

5.

> That's a highly restrictive limitation killing all the benefit of using namespaces."

It is restrictive, but O.K. I think. You do not need two UsersControllers, do you? If so, there is a lot of re-work to do in Cake itself, which nobody wants because it is not needed.

Theoretically, it might be surprizing to see a mini package like this propose things that seem unnecessary or that could be covered by plugins. But in practice, this architecture is simple and the package is non intrusive.

Not your taste?

diegosardina:
Plugin is not always the answer, for this reason in Symfony they relaxed the requirement to organize the code with bundles under src/. Most of the time a subfolder is a simple and straightforward solution.

The main reason is that a plugin/bundle is a piece of reusable code and often the code is so high specialized to a particular software that you will have lot of bundles that strictly depend on the other bundles and the main code. At the end, you can't reuse those plugins/bundles elsewhere.

Developing a true reusable plugin/bundle means to abstract the code more and to use intermediate tables in database to link the various relationships, the latter may introduce a significant overhead.
While this is good thing, I see this good if one really plans to share it among different projects (that may not happen due to the nature of the project).

In my opinion the idea behind this plugin is good but doesn't solve some other problems such having subfolders in the Model folder to avoid very long name of associated models. I have to configure (for every model!) the table locator accordingly for this feature.

Users
User/Data
User/Contacts
User/Addresses
User/Types
User/...

ADmad:
@pabloelcolombiano Organizing one's app using plugins or using folder structure similar to the one supported by your plugin is a matter of personally taste and I don't have issues with either. Please can chose whichever they prefer.

My main issue is with naming the folder groups as "domains" is quite misleading. What you have is just grouping of MVC classes. Your plugin doesn't provide anything which relates to the [building blocks[https://en.wikipedia.org/wiki/Domain-driven_design#Building_blocks] of domain driven design.

Using a term like "module" or "bundle" for your folder groups would be more appropriate.

P.S. In Cake 4 the templates are no longer nested inside the `src` folder, they are in a new `templates` folder at same level as `src`. `src` only contains class files.

dereuromark:
@pabloelcolombiano 
My biggest concern with this plugin is also the issue with templates.
The 3.x way is already in conflict with composer guidelines of only having autoloadable files in src/ autoloader namespace.
So 4.x moves all those to ROOT/templates.
Code that would use your approach would make this harder to very painful to upgrade to 4.x

As for the plugin structure vs your structure.
I think plugins directly in apps provide this very same way by still staying flexible.
This is actually how I always use to separate my domains
See e.e. https://github.com/dereuromark/cakephp-sandbox/tree/master/plugins as a demo showcase.
It has 0 overhead, is basically the same as your approach - but does on top offer the native cake way of having files per type and allowing the 4.x (future) proof way.

ADmad:
TLDR; What you have done is create an alternative folder structure to plugins and given it a different name. IMO it does not provide enough benefits over plugins itself to warrant including it on this curated list. In fact it would create additional work when upgrading your app to 4.x.

pabloelcolombiano:
Thanks for the feedback guys.

I highly praised the fact that templates are moving out *src* in Cake 4.x. I haven't tried it yet, but it sounds really good.

My impression was that this move would make the cake-dm part of jumping from element to element directories accross layers even easier. But I might be wrong.

Regarding the naming in *Domain*, it is indeed ugly, for one because of the name, but even more because this folder is not required. Your reflection on that @ADmad made me reconsider the structure. The *Domain* directory is now gone.

@ADmad yes the layer pattern looks very much like *plugins in src*. But it does not mean that I gave up plugins. It is just that I have a core app (which is really big) located in *src*, and which is shared by all customers, and I have a **** load of satellite functionalities located in plugins. In practice, when I sell my product, I provide the core for free, and each plugin has to be paid extra. This way I always know what my *core* app is. It is also easier to write tests this way.

I have added something to the class name uniqueness in the src folder [here](https://github.com/pabloelcolombiano/cake-dm#optional-1) issue.

@dereuromark I will always recommend splitting an app using the Cake conform approach (plugin based). But for now, refactoring our huge app using plugins is no go. I might not be the only one in this case, that's why I cared about sharing. The cake-dm approach does the job in a day of work, testing included, and I see no security issues if one uses the test described in the link above.

I was surprised to see how smooth the migration went, as well as by the unexpected nice-to-have the concept brought: layout inheritence, depth in the structure.

I would not recommand that to someone learning Cake. But having a second dimension to grow into in addition to plugins is for our app just great. I do not know of any other PHP frameworks that allow such a freedom.

I'll give a try to Cake 4.x with the cake-dm too. I'll fully understand if you rejected the request for now. I might come back with something different. I'll keep you posted!

dereuromark:
All right, lets have a chat again at a later point. Thank you anyway :)

ADmad:
@pabloelcolombiano FYI you can use multiple folders for plugins. So for e.g. you can use `plugins` folder for the ones you included with your app for free and `premium_plugins` for paid ones. Then you just have to update the `App.paths.plugins` config [here](https://github.com/cakephp/app/blob/master/config/app.php#L65) to include the `premium_plugins` folder too.

pabloelcolombiano:
> 
> 
> @pabloelcolombiano FYI you can use multiple folders for plugins. So for e.g. you can use `plugins` folder for the ones you include with your app for free and `premium_plugins` for paid ones. Then you just have to update the `App.paths.plugins` config [here](https://github.com/cakephp/app/blob/master/config/app.php#L65) to include the `premium_plugins` folder too.

@ADmad this is much more powerfull that changing `App.path.templates`. I'll give this one a try, thanks!

ADmad:
@pabloelcolombiano Thank you for taking all the feedback constructively.

PR #337, opened 2020-01-07
Strawberry update

ndm2:
https://github.com/icings/menu


dereuromark:
I dont see the CakePHP 4 requirement: https://github.com/KnpLabs/KnpMenu/blob/master/composer.json#L22

ndm2:
You're looking at the wrong repo, KnpMenu is a dependency :)

https://github.com/icings/menu/blob/4.0.0/composer.json#L17

PR #339, opened 2020-01-07
Adding :strawberry: to Mailgun Plugin

challgren:
https://github.com/narendravaghela/cakephp-mailgun/releases/tag/5.0.0-beta


dereuromark:
I dont think you need min stab dev anymore
https://github.com/narendravaghela/cakephp-mailgun/blob/master/composer.json#L55

PR #352, opened 2020-02-16
Add Ali1/cakephp-brute-force-protection

Ali1:
https://github.com/Ali1/cakephp-brute-force-protection

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,7 @@ Additional lists you might find useful:
 ## Security
 *Plugins and information around security, preventing vulnerabilities and protection against XSS and alike.*
 
+- :strawberry: [Ali1/BruteForceProtectionComponent](https://github.com/Ali1/cakephp-brute-force-protection/tree/master) - Simple way to add Brute Force Protection to your installation without involving database.
 - [BryanCrowe/EncryptedType](https://github.com/bcrowe/cakephp-encrypted-type) - Simple solution that enables Encrypted data types to be stored in the database.
 - :strawberry: [Captcha plugin](https://github.com/dereuromark/cakephp-captcha) - Simple, unobstrusive and extendable captcha solution providing by default an image based math captcha.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -343,6 +343,7 @@ Additional lists you might find useful:
 ## Security
 *Plugins and information around security, preventing vulnerabilities and protection against XSS and alike.*
 
+- :strawberry: [Ali1/BruteForceProtectionComponent](https://github.com/Ali1/cakephp-brute-force-protection/) - Simple way to add Brute Force Protection to your installation without involving database.
```
dereuromark: Doesnt look like this is the actual plugin name. See how to set the proper name as per guidelines.

ADmad:
This would be better suited as a middleware instead of component.

Use the request object instead of PHP globals https://github.com/Ali1/cakephp-bruteforce/blob/0cf218b9bf86f64270f2e7c53248de847811eb19/src/Controller/Component/BruteforceComponent.php#L61

Ali1:
Thanks

More suited but less convenient - let me think about this.

I wanted to do a check on the selected Cache setting, what The ‘duration’ of the config was, but I couldn’t figure it out. What’s the best way?

Ali1:
How do you suggest the middleware is configured. A conditional in Application.php for each $request->getParam('action') that needs brute force protection?

ADmad:
> A conditional in Application.php for each $request->getParam('action') that needs brute force protection?

Only the login page needs protection isn't?  So the middleware can have a config to specify the login URL similar to an authentication middleware.

smarek:
I see value of this plugin, in more places, for example Password Recovery / Forgot Password page.

Ali1:
I'm using it in a number of places within the same application and each one has a different configuration. I'd find it's usage as Middleware difficult.

dereuromark:
A middleware can indeed be quite limiting and hard to configure more flexible based on specific controllers or alike.
As such I would tend you agree with you that the current state is fine.
You could add a later point still add the option of middleware, so people can chose what suites them best.

PR #360, opened 2020-03-24
Add Http/Adapter/Cake library

dereuromark:
https://github.com/php-http/cakephp-adapter
for http://httplug.io/ integration

Required e.g. for using https://github.com/geocoder-php/Geocoder with CakePHP client.

Wasn't totally sure about the architecture headline.
Maybe this could also be sth else?


ADmad:
Yeah putting it under "Architecture" doesn't seem right. We could create a new "HTTP" and put in under that and also move your Geo plugin under it. The current "Geolocation" section could be done away as it's unlikely there will be any under plugin under it.

dereuromark:
I started the HTTP section.
Geo doesn't fit here as it does more than just HTTP, it also has helpers to display maps and calculate distances, custom find('distance') ORM finder etc.

PR #366, opened 2020-05-15
Add ActivityLogger plugin

ishanvyas22:
Add https://github.com/elstc/cakephp-activity-logger

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,7 @@ Additional lists you might find useful:
 ## Auditing / Logging
 *Plugins for auditing and logging.*
 
+- [Activity Logger plugin](https://github.com/elstc/cakephp-activity-logger) - ActivityLogger plugin for CakePHP.
 - [AuditStash plugin](https://github.com/lorenzo/audit-stash) - Flexible and rock solid audit log tracking.
 - :strawberry: [DatabaseLog plugin](https://github.com/dereuromark/CakePHP-DatabaseLog) - Simple and stand-alone logging to database instead of files.
 - :strawberry: [Muffin/Footprint plugin](https://github.com/UseMuffin/Footprint) - Plugin to allow passing currently logged in user to model layer.


dereuromark:
Your plugin name looks like a lie - as per our guidelines
`Elastic/ActivityLogger` seems to be the name.

README.md
```diff
@@ -96,6 +96,7 @@ Additional lists you might find useful:
 ## Auditing / Logging
 *Plugins for auditing and logging.*
 
+- [Activity Logger plugin](https://github.com/elstc/cakephp-activity-logger) - ActivityLogger plugin for CakePHP.
```
dereuromark, line 99: The description could explain a bit more, and the "plugin for CakePHP" can be dropped as this is obvious.

dereuromark:
Code comments in non English is not something that fits our quality criteria for reusability in the community ( https://github.com/elstc/cakephp-activity-logger/blob/master/src/Model/Behavior/LoggerBehavior.php#L217 ).

dereuromark:
In what way is that different/better than the existing AuditStash plugin?
https://github.com/FriendsOfCake/awesome-cakephp#auditing--logging
If this was for 4.x this might be more interesting, but it also seems to be only for 3.x still.

ishanvyas22:
> In what way is that different/better than the existing AuditStash plugin?
> https://github.com/FriendsOfCake/awesome-cakephp#auditing--logging
> If this was for 4.x this might be more interesting, but it also seems to be only for 3.x still.

Yes @dereuromark , you are right. AuditStash plugin is better than ActivityLog, when I looked the AuditStash plugin I was in impression that it will only store data into elastic search(that I don't want). But it also stores activity data into mysql database table, and also well maintained and supports cake 4.x.

So closing the merge request as we don't need new "not-maintained" plugin in awesome list.

PR #369, opened 2020-06-21
FastPaginate

swiffer:
https://github.com/swiffer/FastPaginate


dereuromark:
The plugin seems to be missing some tests
Also the plugin name does not fit what you put into this PR, according to your code, it is swiffer/FastPaginate

swiffer:
Hi @dereuromark - thanks taking a look. could you point my into the right direction on how to best reuse some of the tests written for the built in paginators?

PR #370, opened 2020-07-18
Add plugin skeleton 

ishanvyas22:
Adds https://github.com/ishanvyas22/cakephp-plugin-skeleton

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ Additional lists you might find useful:
 - [Lqdt/OrmJson plugin](https://github.com/liqueurdetoile/cakephp-orm-json) - Behavior and Trait for selecting, finding, getting and setting properties and values inside JSON type fields through CakePHP ORM.
 - [Money plugin](https://github.com/gourmet/money) - Money data type for CakePHP entities using [sebastianbergmann/money](https://github.com/sebastianbergmann/money).
 - :strawberry: [Muffin/Orderly plugin](https://github.com/usemuffin/orderly) - Allows setting default order for your tables.
-- :strawberry: [Muffin/Sti plugin](https://github.com/UseMuffin/Sti) - Single Table Inheritance for CakePHP. 
+- :strawberry: [Muffin/Sti plugin](https://github.com/UseMuffin/Sti) - Single Table Inheritance for CakePHP.
 - :strawberry: [Muffin/Trash plugin](https://github.com/usemuffin/trash) - Soft-delete behavior for CakePHP.
 - [PersistRelatedData plugin](https://github.com/riesenia/persist-related-data) - Behavior for persisting selected fields of related models.
 - [Robotusers/Excel plugin](https://github.com/robotusers/cakephp-excel) - ORM wrapper for PHPExcel.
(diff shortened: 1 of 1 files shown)


dereuromark:
Are there any advantages of such skeletons over normal bake plugin operations?
Especially since most things are hardcoded right now - https://github.com/ishanvyas22/cakephp-plugin-skeleton/blob/master/src/Plugin.php#L41 and would require manual changing over the baked ones.

There is an existing 3.x template one, but that one doesn't provide a composer name yet, so that's a bit more "skeleton" than this one.
Also, you might want to look into https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-template-repository as those could actually make it easier to create one from scratch using GitHub directly.

ishanvyas22:
Hey @dereuromark,

> Are there any advantages of such skeletons over normal bake plugin operations?

Yes. I don't have to install `php-cs-fixer`, `cakephp-codesniffer`, `phpstan`, `phpunit` and other dependencies manually(not to mention the config files each time I have to copy from my past project), also need to add `.editorconfig` file, edit `.gitignore` file to ignore tools/package/etc specific files, etc. 

>  Especially since most things are hardcoded right now - https://github.com/ishanvyas22/cakephp-plugin-skeleton/blob/master/src/Plugin.php#L41 and would require manual changing over the baked ones.

Yes, that things can be dynamic. I can work on that and try to change particular words with actual Plugin name which user has entered as bake does. Please create new issue in the repo, PR also welcome :)


> There is an existing 3.x template one, but that one doesn't provide a composer name yet, so that's a bit more "skeleton" than this one.

IMHO, cloning or downloading zip from repo is not proper way do it. It's like saying install cakephp framework via cloning particular tags or branch. The whole idea of this is, when you go new plugin idea you just use one command to create new package directory and start building it. Also it has `CHANGELOG.md`, `CONTRIBUTING.md` files(which all ideal plugins should have) already ready to go so user just have to remember to create all these stuff every time they want to start something.

> That said: I am not opposed of also having that on in there - as 4.x one.

I might also make this compatible with cake3 too. So I don't think this plugin is only for Cake 4.x.

If you feel anything that needed to change just submit a enhancement request or issue in the repo.

dereuromark:
> IMHO, cloning or downloading zip from repo is not proper way do it

With 10+ years experience in that field I can tell you that's not the case once you actually made enough changes in your project.
Here you can instead use different tools to include changes, e.g. "skeleton" task of https://github.com/dereuromark/upgrade#usage
That would help you, with git diff and some manual merging, to overcome the issue. But trying to resolve hundreds of conflicts from the upstream merges are too painful.

As such, usually, you create your own Git repo from scratch, decoupled from the actual app repo.
But that's just my experience, maybe others see that differently.

ADmad:
Why include `php-cs-fixer` when `phpcbf` can already auto format the code?

cnizzardini:
I will add there is a need for creating plugin skeletons when you are creating stand-alone plugins. I've been working on my own which is still in alpha here: https://github.com/mixerapi/cakephp-plugin-template

Granted that is very specific to my project, but the point is there is an obvious community need for something like this. Whether or not this is the golden ticket...I dunno.

ishanvyas22:
> Why include `php-cs-fixer` when `phpcbf` can already auto format the code?

I don't think it's necessary `phpcbf` does most of the job. I couldn't able find some of the sniffs and PHP CS Fixer was easy to implement so I've added it as a dependency.

But I found the sniff I needed and released a new [version](https://github.com/ishanvyas22/cakephp-plugin-skeleton/releases/tag/v1.2.0) with PHP CS Fixer removed.

ishanvyas22:
@cnizzardini Yeah, as you said I also created this template project myself so I can use it in the future. But I also want to share with the community so all can leverage from this project.

I've checked your template too, look good with coverage report for test and all stuff. I've just a question for you like why are you using `phpmd`? Is there any specific things `phpstan` doesn't do and `phpmd` does?

cnizzardini:
It catches some additional stuff. Is PHPStan capable of running cyclomatic complexity scores? If it is, maybe I don't need PHPMD. But it seems to do some additional stuff.

ADmad:
As @dereuromark pointed out in it's current form with all the hardcoded stuff, I too am not thrilled about promoting this plugin.

You should interactively allow users to update the various details like https://github.com/UseMuffin/Skeleton does for e.g.

ishanvyas22:
Got it @ADmad. Let me implement this and get back to you guys.

ishanvyas22:
Will do a fresh PR with suggested changes. Closing it for now.

PR #381, opened 2020-10-03
Easy Query supports CakePHP4.x

itosho:
- ref: https://github.com/itosho/easy-query/releases/tag/v2.0.0


ADmad:
@itosho The argument type hinting could have been improved in the new major. `EntityInterface` instead of `Entity`, support `iterable` instead of `array` for bulk functions.

itosho:
@ADmad Yes, I get it. Thanks.

PR #383, opened 2020-10-10
Update README.md

EOM:


ADmad:
The master branch is for Cake 3/4 only. Also it's too late to be promoting plugins for Cake 2 which is already EOL.

PR #393, opened 2021-01-14
Mark FractalTransformerView plugin as CakePHP 4 compatible

andrej-griniuk:
https://github.com/andrej-griniuk/cakephp-fractal-transformer-view


ADmad:
The plugin's readme still shows use of special view variables like `_serialize` (which have been deprecated in 4.x) instead of ViewBuilder options.

andrej-griniuk:
> The plugin's readme still shows use of special view variables like `_serialize` (which have been deprecated in 4.x) instead of ViewBuilder options.

Oooops. Will sort it out

andrej-griniuk:
Done!

ADmad:
Your readme example shows setting of paginated articles but I don't see CakePHP specific [paginator adapter](https://fractal.thephpleague.com/pagination/) in the plugin, so your output won't contain any pagination data.

PR #398, opened 2021-05-04
Replace ADmad/HyridAuth with ADmad/SocialAuth

ADmad:


dereuromark:
`https://puppetlabs.com/` seems to be offline

PR #402, opened 2021-06-17
Add icings/partitionable plugin

LordSimal:
Awesome plugin if you need to limit associated data


ADmad:
@ndm2 Do you approve? :)

ndm2:
I mean... generally, yes, but it's not stable yet.

Based on the feedback so far I don't have any breaking changes planned yet, but that could of course change, and I'm not sure if people would appreciate that from something on this list.

ADmad:
I don't think anyone can complain even if you do make breaking changes following semantic versioning.

ndm2:
If you're OK with it, then please, be my guest :)

image

PR #406, opened 2021-06-18
MixerApi/ExceptionRender

cnizzardini:
https://github.com/mixerapi/exception-render


dereuromark:
Those are a lot of very specific, edge case plugins it seems
I wonder if it was better to just list the main one?
Or what's the use case for these small specific ones as standalone usage?

ADmad:
Yeah I too don't see someone using this plugin standalone.

cnizzardini:
I can close this one.

cnizzardini:
> Those are a lot of very specific, edge case plugins it seems
> I wonder if it was better to just list the main one?
> Or what's the use case for these small specific ones as standalone usage?

It's modular components. So it's up to the developer to decide what they need.

PR #407, opened 2021-06-24
Add Galley Plugin

amayer5125:
Galley is a command line utility used to control a small docker environment. It includes bin/cake integration to set up the docker environment on existing applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,7 @@ Additional lists you might find useful:
 - [CakePHP Vagrant Setup](https://github.com/cpierce/cakephp-vagrant-setup) - Tool for spinning up multiple CakePHP 3.x Vanilla Dev Environments.
 - [Devilbox](https://devilbox.readthedocs.io/en/latest/) - A docker development environment for (CakePHP) apps to be auto-setup including a lot of tools.
 - [Docker](https://github.com/stefanvangastel/docker-cakephp) - CakePHP in a docker container environment.
+- :strawberry: [Galley](https://gitlab.com/amayer5125/galley) - A small Docker dev environment for CakePHP development which includes a simple command line utility.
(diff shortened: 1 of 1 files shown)


cnizzardini:
Hey @amayer5125 this is pretty neat. I might want to help out on this. Have you looked at this at all: https://github.com/cnizzardini/cakephp-docker

It does a lot of that stuff, but with a Makefile, in particular it would be cool to see the xdebug toggle integrated into galley. Also the ability to swap out Dockerfiles with userland ones.

amayer5125:
@cnizzardini thanks for your interest! I believe both of your concerns are already solved in Galley.

xdebug can be toggled by adding an environment variable to docker-compose.yml like this:

```yaml
services:
  app:
    environment:
      XDEBUG_MODE: "off"
```

Users can publish the Galley docker files to their repo using `bin/cake galley publish`. This allows them to modify the Galley docker files and rebuild their images. Users can also replace the `build` and `image` config in docker-compose.yml to use any docker image they like.

cnizzardini:
> @cnizzardini thanks for your interest! I believe both of your concerns are already solved in Galley.
> 
> xdebug can be toggled by adding an environment variable to docker-compose.yml like this:
> 
> ```yaml
> services:
>   app:
>     environment:
>       XDEBUG_MODE: "off"
> ```
> 
> Users can publish the Galley docker files to their repo using `bin/cake galley publish`. This allows them to modify the Galley docker files and rebuild their images. Users can also replace the `build` and `image` config in docker-compose.yml to use any docker image they like.

Yeah but you can simplify for the user's sake by modifying the compose file via sed or whatever, then triggering a restart of the php container.

amayer5125:
Do users enable/disable xdebug that often? I could see the use case where a user doesn't want XDebug at all will disable it, but I don't believe I've seen people turn it on and off throughout the day. I could be in the minority here though.

PR #410, opened 2021-10-17
Add FormBuilder library

bancer:
https://github.com/bancer/form-builder

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ Additional lists you might find useful:
 - [DatabaseBackup plugin](https://github.com/mirko-pagliai/cakephp-database-backup) - A plugin to export, import and manage database backups.
 - :strawberry: [Feedback plugin](https://github.com/dereuromark/cakephp-feedback) - Allow visitors to send quick and easy feedback incl. a screenshot via sidebar form.
 - :strawberry: [Flash plugin](https://github.com/dereuromark/cakephp-flash) - More powerful flash messages for your application.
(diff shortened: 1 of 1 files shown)


dereuromark:
At this point we only accept new plugins that can be 4.x compatible

bancer:
I have not tested it with 4.x yet but I believe it is compatible.

bancer:
I did a quick testing with CakePHP 4.2.10. It is compatible.

ADmad:
> I did a quick testing with CakePHP 4.2.10. It is compatible.

Sorry, that's not good enough. You need to run the testsuite with CakePHP 4 as part of your CI setup.

bancer:
> Sorry, that's not good enough. You need to run the testsuite with CakePHP 4 as part of your CI setup.

That would mean that many assertions would need to be rewritten because 4.x FormHelper generates different HTML than in 3.x. I really see no good reason to do that because the code of FormBulder itself will not need to be changed in order to pass the new tests. Take a look at this method for example: https://github.com/bancer/form-builder/blob/master/src/DateTag.php#L68-L72
```
    public function __toString()
    {
        $options = $this->getOptions();
        return $this->FormHelper->date($this->fieldName, $options);
    }
```
This is compatible both with 3.x and 4.x although would generate different HTML as expected.

dereuromark:
You need to proof your compatibility by actuall using require on all major versions and provide a test for highest and lowest (prefer lowest) as we always do.
This way, it can be sure not to break in the future somehow for older ones (hidden regressions).

If outputted HTML is different, maybe assert on it more dynamically, or write a test case per major, and skip for non fitting major.
There are quite a few ways to do it.

Usually however, you would just support 1 framework major per branch/major, and just merge across them.

PR #414, opened 2022-10-09
Adding LogReader

zunnu:
https://github.com/zunnu/log-reader

Log reader makes it possible to read the logs under the /logs folder from the UI. It is also possible to create your own log reading / parsing application using the API.

For example if you have multiple CakePHP installations and you need to monitor the logs folder logging. With this plugin you can create an external system and  show or parse the logs using the API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,7 +343,7 @@ Additional lists you might find useful:
 - [CrudJsonApi plugin](https://github.com/FriendsOfCake/crud-json-api) - Crud listener for building [JSON API](https://jsonapi.org/) compliant APIs.
 - :strawberry: [FractalTransformerView plugin](https://github.com/andrej-griniuk/cakephp-fractal-transformer-view) - A plugin which allows using [Fractal transformers](https://fractal.thephpleague.com/transformers/) for your API output.
 - :strawberry: [MixerApi](https://mixerapi.com) - Streamline development of modern RESTful APIs for your teams CakePHP project.
-- :strawberry: [SwaggerBake plugin](https://github.com/cnizzardini/cakephp-swagger-bake) - This plugin automatically builds your Swagger UI documentation from your existing models and routes. A redoc option is also available.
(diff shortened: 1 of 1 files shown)


dereuromark:
So is this only for local dev?
As on prod systems, you should never use the file system for logging, ideally you use DB or some service logging.

As for the "external" access, this seems also unsafe and quite resource-wasteful.
The above approach also solves that cleanly across a multi server setup IMO.

I am not sure we should promote this here.
Maybe I am missing sth
cc @ADmad

zunnu:
This is for production and dev.
For the API I plan to add condition where the debug mode needs to be on in order for the API to work to improve safety.

I personally don't log errors to database this of course depends on the nature of the application.
There are other ways to do this kind of logging some users use Sentry other use Papertrail.

We often do only CLI based apps meaning our app talks to 3-party apis that can be rest, soap etc.
We have a internal system in use that setups the servers / instances and is able to change the settings of an instance.
Errors like Soap connection failed are checked on specific intervals and if the connection problem is prolonged our customer team informs the customer that there is a problem with the service and our technical team is looking into it.

This is just a one way to do it I agree its resource-wasteful there's no question about it, but I also think it's a good thing people can build their own external logging systems if need be.
In my opinion the framework should also not too strictly say how logging is done. Saying that in production systems the logs folder should not be used at all for logging and that logs should be stored in database is a bit strict IMO. Ideally you would log into syslog ofc.

ADmad:
@zunnu You understand your usecase well and using file based logs might be good enough in spite of it's short coming.

But many others might end up using it where it's not a right fit and potentially create a security risk if used for web apps in production. So I too am vary of promoting it.

zunnu:
@ADmad, @dereuromark Thank you for the feedback. I think I will add a disclaimer to the readME that some kind of request based authentication should be used with this in terms of using the UI. I can also add some information how logging should be done in production so that we can avoid people using this wrong. Would you happen to have articles related to how logging should be done in cake production?

dereuromark:
I am not sure we can move forward given the possibly security issues present by default

At the very least this should be mentioned and used only in dev mode (localy):

    composer require-dev zunnu/log-reader

Having prod logs via URL available that easily seems dangerous none the less right now.

PR #416, opened 2022-10-25
add lordsimal/cakephp-sentry

LordSimal:
I didn't remove the plugin from Connehito because it is still the recommended one for CakePHP 4.0 till 4.4
Since the plugin author didn't respond to my PR I created my own package.


ADmad:
Since your plugin is Cake 4.4+ you can rename the `Plugin` class to `CakeSentryPlugin` instead.

Wondering if we can just drop the original plugin from the list if you add some info in your readme to use that for older Cake versions. Having both listed here could be confusing to users.

LordSimal:
I made some more adjustments to make the plugin also fit the official Sentry PHP SDK requirements (not throw exception if no DSN is set) and I added an additional info block at the top of the Readme to mention the Cake 3.x and 4.0-4.3 version from Connehito.

PR #418, opened 2022-10-27
Add official Content Management Tutorial

ishanvyas22:
Since 4.x is current stable version of CakePHP, we should direct readers to the current version's tutorial page.

I've kept blog tutorial because we still have support for CakePHP 3.x. IMO, we keep it here for now and once we end support for CakePHP 3.x, then we can remove it. Wondering what core members thinks?


ADmad:
Security fixes for 3.x ends 15th December 2022, so not far away :)

PR #419, opened 2023-01-10
Init Cake4+

dereuromark:
We should set this to master now

Before, we need to remove all strawberries, as we will set a new icon for Cake5+

Note: I moved the current content of master into a cake3 branch already.


dereuromark:
With https://github.com/cakephp/cakephp/releases/tag/5.0.0-beta1 already released, should we merge this now?
Or wait till RC?

PR #423, opened 2023-10-05
Add Icon to Denote Galley CakePHP 5 Support

amayer5125:
As of version 1.5.0 [Galley](https://gitlab.com/amayer5125/galley) works with CakePHP 5.x applications.


dereuromark:
You should not be using this kind of constraints IMO:
https://gitlab.com/amayer5125/galley/-/blob/1.5.0/composer.json?ref_type=tags#L14-15

amayer5125:
@dereuromark I don't believe I set those manually. They may be the defaults from when I `bin/cake bake`ed the plugin. Do you have a recommendation for what they should be?

dereuromark:
Do you have extensive test coverage for different cakephp major versions?
Usually you major also your plugin in each cake version and maintain them separately with ^3.x vs ^4.x etc instead of >=
as this can easily lead to BC breaks or broken support.

PR #456, opened 2024-10-29
Add back ApiPagination and EncryptedType plugins

bcrowe:
They were removed in an initial clean up of non-compatibility with 4.x, but since then they've been updated to be compatible with 4.x for a while now. They'll both soon be getting updated for 5.x compatibility as well.


dereuromark:
This late in the 5.x timeline I would imagine that only plugins that can be 5.1+ compatible should be re-added.
But since we dont have a 4.x branch, and have them combined, maybe this is OK for now.

ADmad:
> But since we dont have a 4.x branch, and have them combined, maybe this is OK for now.

With the release of Cake 5.1, I think it's time to make a 4.x branch and drop the ones that haven't been updated from `master`.

dereuromark:
Let's merge this in and then  start to branch off here maybe now?

PR #465, opened 2025-08-09
Add CakeUid plugin

josbeir:
https://github.com/josbeir/cakephp-uid


dereuromark:
Your release is a bit chaotic, it is not clear which is the actual version.

josbeir:
Packagegist didn't pick up the 1.0 tag, was because a missing psr-4 key for dev deps in my composer file I didn't notice. 

Bumping fixed it. Not sure why versioning is a concern here though.

PR #467, opened 2025-08-20
Database-driven message queue integration for CakePHP using the Enqueue library for CakePHP Queue plugin.

skie:


dereuromark:
isnt the plugin name `Cake/Enqueue` as per composer?

dereuromark:
See CI fails.

PR #469, opened 2025-08-20
A CakePHP plugin that seamlessly transforms server-side validation rules into client-side JSON Logic validation, providing automatic form validation without requiring separate client-side validation code.

skie:
https://github.com/skie/rule-flow


dereuromark:
Readme says CakePHP: 4.5+, but is 5.1+ by the way.
Better to use a wiki matrix, see my plugins for better up to date info across branches.

PR #472, opened 2025-09-08
Add TemplaterDefaults plugin

josbeir:
See https://github.com/josbeir/cakephp-templater-defaults

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ Additional lists you might find useful:
 - [Feed plugin](https://github.com/dereuromark/cakephp-feed) - Containing an RssView class to easily generate (complex) RSS feeds.
 - [LatteView plugin](https://github.com/josbeir/cakephp-latte-view) - A plugin providing Latte template engine integration.
 - [Meta plugin](https://github.com/dereuromark/cakephp-meta) - Makes handling meta tags and SEO-relevant HTML markup DRY and easy.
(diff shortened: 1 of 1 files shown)


josbeir:
> Going through the readme examples:
> 
> ```
> <?= $this->Form->control('field', ['class' => 'false']); ?>
> // <input type="text" name="field" />
> ```
> 
> Shouldn't that be boolean `false` instead of string `'false'`?
> 

Should indeed be a string, parsing is happening after attributes are parsed with `formatAttributes` which strips the value.

> ```
> <?= $this->Form->control('field', ['class' => '!!there-can-be-only-one']); ?>
> // <input type="text" name="field" class="there-can-be-only-one" />
> ```
> 
> So the `!!` is a "magic" operator?

Correct, if you want to keep your runtime class but not the classes set in the defaults property (as an array) you would add !! before your property to keep it.

- Attributes set as array will be merged with ones from runtime
- Attributes set as string will be replaced

A side node:

The whole idea works around the current limitation on StringTemplate where there is not enough context in the formatAttributes part (there's no way to determine for what template we are formatting attributes). I know there is some [conversion](https://github.com/cakephp/cakephp/issues/18088) about this, but in the meantime this could be  working and simple enough solution.

ADmad (approved):

PR #475, opened 2025-09-19
cakephp enqueue database transport

skie:
https://github.com/CakeDC/cakephp-enqueue


dereuromark:
Maybe check the CI fails for relevance.

PR #508, opened 2026-08-25
Add Heartbeat plugin

Neluxx:
Adds Heartbeat, a CakePHP plugin providing an application heartbeat status page with configurable sensors.

https://github.com/orca-services/cakephp-heartbeat


dereuromark:
I guess this is very similar to existing
https://github.com/dereuromark/cakephp-setup/blob/master/docs/healthcheck/index.md

But fair enough I guess :)

